PR #3, opened 2024-07-22
Add combined swing mode for horizontal and vertical support

Pr0mises:
This adds the possibility to change the horizontal and vertical swing mode of the AC

### DIFF
--- a/rethink/cloud/devices/RAC_056905_WW.js
+++ b/rethink/cloud/devices/RAC_056905_WW.js
@@ -28,6 +28,10 @@ class Device extends HADevice {
 			},
 			write_xform: (val) => {
 				const modes2clip = { cool: 0, dry: 1, fan_only:2, heat:4, auto:6 }
+				if (val === 'off') {
+		                    // Call function power (0x1f7) with value OFF
+		                    this.setProperty('power', 'OFF')
+		                }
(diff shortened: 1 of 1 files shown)


anszom:
I'm not sure if this is a good idea. The "combined" property only allows HA to "see" the last attribute that it has changed, but not the other one (for example: if you set "Vertical 2", you can't check what is the horizontal swing mode).
- the "right solution" would be to pester HA maintainers to add the other swing mode attribute upstream. 
- as a theoretical alternative, we could have a combined Horizontal x Vertical attribute with 8*9=72 possible values. This - obviously - would not be very practical.
- the current solution: we expose the other swing mode on MQTT, although in a way which is not directly compatible with current HA.

So far the last option seems the most acceptable to me

Pr0mises:
I was thinking about those options but that's how the thingq hacs Integration dealt with it and it wasn't confusing at all. For my part I leave the vertical swing at one position most of the time

Pr0mises:
Good news:
https://developers.home-assistant.io/blog/2024/12/03/climate-horizontal-swing/

Horizontal swings are now supported

anszom:
Sorry, this "fell between the cracks" at some point. Superseded by #36.

PR #4, opened 2024-07-22
Add comments to Device class

Pr0mises:
This adds comments to base.js and RAC_056905_WW.js

I wanted to do multiple PRs for the ease of allowing and declining certain changes


anszom:
I like the comments, but I have two issues with the PR.
First, **please** do not reformat the code. Tab indentation may not be the most common nowadays, but I'm going to stick with it :)
Second, I'm confused about the reverts. My *guess* is that you wanted to keep each change separate, not undo your changes...

Pr0mises:
Ah sorry about that, I missed the spaces change. Looks like VScode did an auto format.

Yes I wanted to make them separate and used the wrong branch. Wasn't sure if it's possible like that.

If not I'll make the changes next week

Drafteed (commented):
I think this can be closed, as #36 has significantly changed the code.

PR #7, opened 2024-09-07
Fix typo and ca file reference

zivillian:


anszom:
Thanks! I wonder how I missed these...

PR #13, opened 2025-11-20
Basic Dockerfile and Compose Setup as a Starting Point

moutasem1989:
After some experimentation, I finally managed to build an image that stays running instead of crashing on startup. I used `ubuntu:24.04` as the base image — definitely heavier than ideal, but it worked reliably — and set `rethink-cloud.js` as the image entry point. I'm still not entirely sure which files need to persist, but for now I’ve mounted only the configuration file `/rethink/config.json` into the container. With that in place, the output behaves exactly as described in the [wiki](https://github.com/anszom/rethink/wiki/Installing-rethink%E2%80%90cloud#step-4-start-the-server):

> root@homeserver:/services/rethink# docker compose logs -f rethink rethink  | Rethink cloud ready
> rethink  | During setup, please ensure that connections to common.lgthinq.com:443 are redirected to lgthinq.mydomain.tld:443
> rethink  | HA mqtt connection established

In my case i created a DNS Record for both `common.lgthinq.com` and `lgthinq.mydomain.tld` pointing to my server with NginX. There I redirected one domain to the other. Is it posible in this case to simply use port 443 and let NginX handle the certificates ?

I have not yet tried it with my ThinQ machine because I do not have a wifi enabled device where i can run `rethink-setup.js` So the image is not yet tested.
This byfar not a production ready solution but it is just a start to get things working

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,39 +1,22 @@
-# Use a small, recent Ubuntu LTS
-FROM ubuntu:24.04
+FROM node:trixie
 
-# Avoid interactive prompts during package installs
-ENV DEBIAN_FRONTEND=noninteractive
-ENV NODE_ENV=production
-
-# Install nodejs and npm via apt (silent)
-# Note: Ubuntu repo versions may be older; if you want newer Node, consider NodeSource or nvm.
-RUN apt-get update \
- && apt-get install -y --no-install-recommends \
-      ca-certificates \
-      curl \
-      gnupg \
-      nodejs \
-      npm \
- && rm -rf /var/lib/apt/lists/*
-
-# Create app directory and set as working dir
+# Create app directory
 WORKDIR /rethink
 
-# Copy application files into the image
-# Make sure your build context contains the ./rethink folder
+# Copy project
 COPY ./rethink /rethink
 
-# Install dependencies at build time.
-# Prefer npm ci when package-lock.json is present for deterministic installs,
-# otherwise fall back to npm install.
-RUN if [ -f package-lock.json ]; then \
-      npm ci --only=production; \
-    else \
-      npm install --production; \
-    fi
+# Install dependencies
+RUN npm install
 
-# Expose port(s) if your app uses any (adjust as needed)
+# Copy entrypoint script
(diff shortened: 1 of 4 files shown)


moutasem1989:
I’ve made major improvements, and I can say it’s now ready for testing:

I created a `Dockerfile` that installs the project into a `node` Docker image and sets up the `entrypoint` correctly.
The entrypoint generates the configuration file based on environment variables and links the files from `/config` into the working directory `/rethink` inside the container (including bridge files and certificates - it rewrites what ever is there).

Depending on the `RETHINK_SERVER_MODE`, it then starts either the cloud or bridge services (with cloud).
The cloud runs as its own service, while the bridge service is launched by a script that monitors the cloud service logs. For every new device that connects (new device ID), it spins up a dedicated bridge service instance and ensures there are no duplicates. it collects all the needed informations accordingly.

I think the only inconvenience is provisioning new devices in the cloud/bridge services. Additionally, we might need a way to automatically copy the `.bridge-{device-id}.json` files—once they’re created in `/rethink`—into `/config` so they persist between container reboots.

The certificate generated by the cloud service is copied into `/config`, but it’s not overwritten; if the domain changes or certificate value changes, it must be removed manually.

I hope this helps. It was a great learning experience for me to build all of this.

Drafteed:
This works. 

However there is one issue. Currently the path to `config.json` is hardcoded. Because of that, mounting the `/data/` directory with configs/certificates inside has no effect:

**docker-compose:**
```yaml
services:
  rethink:
    container_name: rethink
    image: rethink
    restart: unless-stopped
    volumes:
      - ./data:/app/data
    network_mode: host
```

For this to work, `config.json` and each cert file need to be mounted into the app root directory:

```yaml
services:
  rethink:
    container_name: rethink
    image: rethink
    restart: unless-stopped
    volumes:
      - ./data/config.json:/app/config.json
      - ./data/ca.cert:/app/ca.cert
      - ./data/ca.key:/app/ca.key
    network_mode: host
```

Ideally, all configuration files, certificates, and state files should be moved to the `data` directory (in the project).

anszom:
a different approach was merged, closing

Issue #25, opened 2026-03-11
Project proposals


Drafteed:
@anszom You've created a great project!

I have a few suggestions:

1. ~Add **Prettier** + **ESLint** to the project, or alternatively use all-in-one **Biome**.~
2. ~Add **Husky** or **simple-git-hooks** alongside the tools above so commits are automatically checked/formatted.~
3. ~Also add CI (GitHub workflow) for the tools above to check code and formatting in PRs.~
4. ~Create a `dev` branch and use it for development. Keep `master` as the stable branch containing only the latest stable release.~
5. It might also be worth adding the ability to customize devices in the configuration - for example, enabling/disabling/override certain sensors and renaming devices.
6. I would also suggest moving the frontend into a separate npm package in the future and refactoring it, for example using React.

anszom:
> The implementation way reminds of Valetudo, which after a while this project may also evolve into something like that where instead of deploying stuff on your network, you simply(easier said than done) build a custom binary that contains the original fw with dns patches, rethink-cloud, a web ui and/or mqtt. Something that can be flashed to the module itself.

I'm also a valetudo user so I get the appeal, but this isn't really feasible here. Valetudo targets devices that already run an embedded Linux system on relatively powerful hardware. Thinq devices run on hardware which is very limited compared to this (for an analogy, compare raspberry pi vs esp32). Re-flashing the original firmware would require significant reverse-engineering effort, be way more invasive, have a potential of bricking the modem, etc. And you would still be expected to run HomeAssistant somewhere.

maciejsszmigiero:
Progress update:
<img width="432" height="1074" alt="additional controls" src="https://github.com/user-attachments/assets/c87c0dde-5ef8-401c-bd00-741bc874f9ad" />

Still working on it, but getting closer to patches being ready.

maciejsszmigiero:
Pull request for the above improvements now opened as #36

Drafteed:
@anszom I noticed an interesting thing. For ThinQ2 devices, it's not necessary to go through the provisioning process via rethink. It's enough to start adding the device in the official LG app, then add a DNS entry on the router pointing `common.thinq.com` to the local server, and complete the setup - the device will be successfully added to rethink. After that, `common.thinq.com` DNS record can be removed if you want to use bridge mode.

Thus, the setup process can be described as follows:

1. Deploy rethink via Docker or as an npm package;
2. Edit the config: set `hostname = rethink.lgthinq.com` and port `443`, then start rethink;
[...]

PR #36, opened 2026-04-17
RAC_056905_WW additional functionality

### LINKED ISSUE #19: RAC_056905_WW Jet Mode
I was reversing here to find out how to enable Jet mode from the air conditioning. And it its at reg 0x323. Its default to 0 and setting it to 1 while in `cool` mode triggers it. That makes the fan speed faster than normally allowed and I think the compressor also runs at higher power (I can clearly see it consumes about the maximum rated power, which I cant make it happen with other modes).

It **needs** to be in Cooling mode to be activated, otherwise it just ignores it. So I ended up doing this:

```ts
                this.addField(config, {
                        id: 0x1f9, name: 'mode',
                        read_xform: (raw) => {
                                const modes2ha = [ 'cool', 'dry', 'fan_only', undefined, 'heat', undefined, 'auto', 'jet' ]
                                if(this.raw_clip_state[0x1f7] === 0)
                                        return 'off'
                                if(raw == 0 && this.raw_clip_state[0x323] == 1)
                                        return 'jet'
                                return modes2ha[raw]
                        },
                        write_xform: (val) => {
                                const modes2clip = { cool: 0, dry: 1, fan_only:2, heat:4, auto:6 }
                                if (val === 'off') {
                                    this.setProperty('power', 'OFF')
                                }
                                if (val === 'jet') {
                                    this.setProperty('jet_mode', 'on');
                                    this.setProperty('swing_modes', '0');
                                    return 0; // Cool
                                } else {
                                    this.setProperty('jet_mode', 'off');
                                }
                                return modes2clip[val]
                        },
                        write_attach: [0x1fa, 0x1fe, 0x323, 0x322]
                })

                this.addField(config, {
                        id: 0x323, name: 'jet_mode',
                        read_xform: (raw) => {
                                const modes2ha = [ "off", "on" ]
                                return modes2ha[raw]
                        },
                        write_xform: (val) => {
                                const modes2clip = { "off": 0, "on": 1 }
                                if (val === 'on') {
                                        this.setProperty('mode', 'cool'); // Jet Mode requires cool mode
                                        this.setProperty('swing_modes', '0');
                                }
                                return modes2clip[val]
                        },
                        write_attach: [0x1f9, 0x1fa, 0x322]
                })
```

Which works to toggle on/off by setting mode through MQTT to `jet`. Swing Modes is optional, but setting it to 0 makes it almost fully 90 degrees (which is what it does when you hit the JET MODE button in the controller).

Sadly I didnt find a way to integrate at home assistant on the `climate` entity. the HVAC modes are fixed there, and I'm not sure how to add it as a "new button".

### LINKED ISSUE #29: req_timesync
I'm getting a lot of messages like these:

```
2026-03-25T15:02:52.263Z incoming clip/message/devices/<redacted> {"mid":114424299,"did":"<redacted>","kind":"F_V8_Y___W.B_2QEUK","cmd":"req_timesync","rssi":-47,"ble_rssi":-100,"fs":"idle","br":4800,"per":-1,"data":null,"type":1}
Mar 25 16:03:02 rethink npm[273]: 2026-03-25T15:03:02.315Z incoming clip/message/devices/<redacted> {"mid":114434351,"did":"<redacted>","kind":"F_V8_Y___W.B_2QEUK","cmd":"req_timesync","rssi":-47,"ble_rssi":-100,"fs":"idle","br":4800,"per":-1,"data":null,"type":1}
Mar 25 16:03:12 rethink npm[273]: 2026-03-25T15:03:12.319Z incoming clip/message/devices/<redacted> {"mid":114444355,"did":"<redacted>","kind":"F_V8_Y___W.B_2QEUK","cmd":"req_timesync","rssi":-47,"ble_rssi":-100,"fs":"idle","br":4800,"per":-1,"data":null,"type":1}
```
My IoT-Wifi doesn't allow communications to wan. Could this be the machine trying to access an outside timeserver?
My DHCP offers a timeserver and it is reachable from within my IoT-Wifi.
My DNS only shows the machine trying to resolve rethink.
Thoughts?

maciejsszmigiero:
This pull request implements quite a lot of additional functionality to `RAC_056905_WW` devices - as announced in #25 some time ago.

* Both horizontal and vertical swing configuration for the Climate component (replacement for #3),
* Jet cool and jet heat (fixes #19),
* Air cleaning ("Plasma"),
* Energy saving mode,
* Turn-on, turn-off, sleep timers,
* Auto dry switch and auto dry remaining time monitor,
* Diagnostic values like error code, EEV opening, IDU capacity,
* Time sync response so modems no longer keep pooling for time every minute (fixes #29),
* Complete basic filter management.
    
Each of these functionalities should be auto-detected based on the device capability query response.
Note that this might cause  Home Assistant to re-detect the `climate` component of  `RAC_056905_WW` devices since these devices are now `device` component based.

Also included are a few minor additional changes, like small debug improvements or a new `packet-sender-device` tool.

Please refer to individual commit messages for details of these changes.


maciejsszmigiero:
Looks like GitHub change viewer messed up indentation a bit by converting spaces into tabs (or vice versa) which does not happen under raw `git`.

Drafteed (commented):
Consider splitting the PR into several smaller ones to simplify the review process.

maciejsszmigiero:
> Consider splitting the PR into several smaller ones to simplify the review process.

The patch set is already split in 13 commits.

Each commit can be reviewed individually.

volleybus:
@maciejsszmigiero can confirm most of the new functionalities work as intended. The optional features have not kicked in for me although some were available in the LG app such as current power usage (may be an approximation based on usage or my module reports on a different offset?). I might be able to check my module responses to match a pattern but all I can see is the code checks if a response is received at 0x32e. Same goes for error and valve values.

Thanks for your work!

rethink/cloud/devices/tlv_device.ts
```diff
@@ -49,10 +51,29 @@ export default class TLVDevice extends HADevice {
...
     // clip-side
-    start() {
+    query() {
         this.send([1,1,2,2,1], [{t: 0x1f5, v: 2 }])
     }
 
+    start() {
+	this.query()
+
+	// Refresh every 15 minutes since not every tag change generates async notify
```
anszom, line 93: I'm curious - what kind of tags are these? And how does the official backend handle these then...

rethink/cloud/thinq2/device.ts
```diff
@@ -23,11 +24,15 @@ export class Device extends TypedEmitter<DeviceEvents> {
 		super()
 	}
 
-	send(buf: Buffer) {
-		this.emit('sendData', buf)
-		const messagestr = JSON.stringify({"did": this.id,"mid":Date.now(),"cmd":"packet","type":1,"data":Buffer.from(buf).toString("hex")})
+	send(cmd: string, type: number, data: string | JSONObject) {
+		const messagestr = JSON.stringify({"did": this.id,"mid":Date.now(),"cmd":cmd,"type":type,"data":data})
 		this.broker.publish({topic: this.topic, retain: false, qos: 0, dup: false, payload: messagestr}, null /* message generated by the broker itself */)
 	}
```
anszom, line 30: > Some of future functionalities require sending device more than just a
"packet" command with type of 1

Do you have any examples? I couldn't find these in your PR.

If so, then we will likely need to expand the bridge part as well in order to handle these packets coming in from the LG cloud...

rethink/util/util.ts
```diff
@@ -2,3 +2,8 @@
 export function allowExtendedType<T, Q extends T>(t: Q): T {
     return t;
 }
+
+type JSONScalar = string | number | boolean | null;
+export type JSONValue = JSONScalar | JSONObject | JSONArray;
+export type JSONObject = { [k: string]: JSONValue };
+type JSONArray = JSONValue[];
```
anszom: Let's simply use a typescript `object` like with thinq1

rethink/cloud/devices/tlv_device.ts
```diff
@@ -110,8 +130,33 @@ export default class TLVDevice extends HADevice {
         this.thinq.send_packet(Buffer.from(buf))
     }
 
+    capabilityKeyTry(k: number) {
+	/* To be overridden */
+	return false;
+    }
+
+    capabilityReceived() {
+	throw new Error("To be overridden");
```
anszom: The implementation in WIN_056905 proves that this can be safely left empty, so let's remove the throw and leave the default implementation a no-op.

rethink/cloud/devices/WIN_056905_WW.ts
```diff
@@ -127,4 +127,11 @@ export default class Device extends TLVDevice {
 
     this.setConfig(config)
   }
-}
\ No newline at end of file
+
+  capabilityKeyTry(k: number) {
+    return k == 0x2da /* eeprom checksum */
+  }
```
anszom: I think something like this would be clearer:
```
isCapsResponse(tlvArray: TLV.TLV[]) {
    return tlvArray.any(({t, v}) =>  t === 0x2da)
}
```

rethink/cloud/homeassistant.ts
```diff
@@ -21,7 +21,8 @@ import log from '../util/logging.js'
 function recursiveReplace(obj: unknown, replacements: Record<string, string>) {
 	if(Array.isArray(obj)) {
 		return obj.map((v) => recursiveReplace(v, replacements))
-
+	} else if(obj === null) {
+		return null
```
anszom, line 25: nitpick: these changes should be moved into a different commit

rethink/cloud/homeassistant.ts
```diff
@@ -171,7 +172,7 @@ export type AvailabilityInfo = {
 }
 
 export type ComponentInfo = {
-	name?: string
+	name?: string | null
```
anszom, line 175: nitpick: these changes should be moved into a different commit

anszom (changes requested):
First of all: this is an impressive PR. I can see why it took you that long to finalize it. There are several small issues that I'd like fixed, see my comments.

The formatting issue is real, not a github artifact. The project had started with only hard 8-space tabs, then it got a bit inconsistent during the typescript port. Still, within a file, I'd like to stick to one option, so if a file uses tabs, please keep your changes tab-formatted as well.

Also, I'm curious - where did you get such detailed information about these devices? Any hidden docs that I should be aware of? :) Don't tell me you guessed all of this :P

maciejsszmigiero:
> @maciejsszmigiero can confirm most of the new functionalities work as intended. 

Lovely, thanks for testing!

> The optional features have not kicked in for me although some were available in the LG app such as current power usage (may be an approximation based on usage or my module reports on a different offset?).

The trouble here is that I have a _Multi_ system and these do *not* report their energy or power usage via ThinQ or in the app, even though ODU does report its instantaneous power consumption via _LGMV_ interface.
So unfortunately  I have no hardware to develop and test such energy reporting functionality against.

> I might be able to check my module responses to match a pattern but all I can see is the code checks if a response is received at 0x32e. 

Tag _0x32e_ is actually the nominal IDU cooling power capacity, it is *not* the power consumption.

anszom:
Power consumption should be available in field 0x232, although I've never gotten to decoding its units

maciejsszmigiero:
> The project had started with only hard 8-space tabs, then it got a bit inconsistent during the typescript port. Still, within a file, I'd like to stick to one option, so if a file uses tabs, please keep your changes tab-formatted as well.

Sure, but what style is preferred for mixed files like `tlv_device.ts`? This file is essentially half-tabs, half 4 spaces indention.

> Also, I'm curious - where did you get such detailed information about these devices? Any hidden docs that I should be aware of? :) Don't tell me you guessed all of this :P

Just good ol' reverse engineering from Thinq API, modem firmware, LGMV values...

> Power consumption should be available in field 0x232

Unfortunately, this tag is always 0 on every IDU for Multi devices so I can't really help here.

rethink/cloud/devices/tlv_device.ts
maciejsszmigiero: I think at least the EEV opening one at 0x330, not sure about the error code one at 0x221.

It seems to me the official backend also does some periodic queries.

In general, these devices are sometimes rather erratic - for example the "WiFi waves" indicator light sometimes goes out on an IDU for no obvious reason, even though the device remains connected to MQTT and responding to commands.
So doing a periodic re-query also helps mitigate lost tag updates there.

rethink/cloud/thinq2/device.ts
maciejsszmigiero: Yeah, look for the `setMaskingInfo` command in the `capabilityReceived()` handler.

anszom:
> Sure, but what style is preferred for mixed files like tlv_device.ts? This file is essentially half-tabs, half 4 spaces indention.

Ok, you're right it's a mess. Keep your formatting as-is. #42 will clear that up, but let's merge this one first.

rethink/cloud/thinq2/device.ts
anszom: Ok I see. I'll take a look at the bridge side after we merge this.

skinkie:
UVnano:
t= 2a2 v= 0, t= 2a2 v= 1

maciejsszmigiero:
Updated the PR with the requested changes.

Reformatted the new code in this PR to preserve the neighboring whitespace style where possible.

Drafteed:
My parents now have this kind of AC, so I'll can to test this change tomorrow when I visit them.

rethink/cloud/devices/RAC_056905_WW.ts
```diff
@@ -66,40 +248,400 @@ export default class Device extends TLVDevice {
...
+			config['components']['filterlife'] = filterLife
+			const filterChanged = {
+				platform: 'sensor',
+				unique_id: '$deviceid-filterchangeddate',
+				state_topic: '$this/filterchangeddate',
+				name: 'Filter usage last reset',
+				icon: 'mdi:calendar-refresh-outline',
+				device_class: 'date',
+				entity_category: 'diagnostic',
+			}
```
Drafteed, line 404: Got a warning in logs:
```
Invalid state message '0000-00-00' from 'rethink/XXXXXXXXXXXX/filterchangeddate'
```

I think we need to add an additional check or remove device_class: 'date'.

rethink/cloud/thinq2/device.ts
```diff
@@ -23,11 +23,17 @@ export class Device extends TypedEmitter<DeviceEvents> {
 		super()
 	}
 
-	send(buf: Buffer) {
-		this.emit('sendData', buf)
-		const messagestr = JSON.stringify({"did": this.id,"mid":Date.now(),"cmd":"packet","type":1,"data":Buffer.from(buf).toString("hex")})
+	send(cmd: string, type: number, data: string | object) {
+		const messagestr = JSON.stringify({"did": this.id, "mid":Date.now(),
+			"cmd":cmd, "type":type,
+			"data": (typeof(data) === 'string' ? data : { ...data }) })
```
anszom: Why not simply pass `data` as-is ?

rethink/cloud/devices/RAC_056905_WW.ts
maciejsszmigiero: If the filter change date has never been (re)set, then the device reports last filter change date of `0000-00-00`.

HA does not like that date and does not seem to accept any other special value like `null` to make a `date` sensor report as "Unknown".
Rewriting `0000-00-00` to some other magic arbitrary date like `1970-01-01` instead would be a bit misleading.

Considering that the HA warning is otherwise harmless and will be gone forever after the first filter reset I think adding complexity with custom availability topic for that single sensor or reducing metadata accuracy by turning this sensor data type to free text is not warranted.

PR #42, opened 2026-04-20
Add Prettier to the project

Drafteed:
This PR introduces Prettier to the project and reformats the existing codebase to keep styling consistent. It also adds a GitHub Actions CI workflow to automatically run formatting and build checks for new changes. This also adds Git hooks to automatically reformat before pushing.

No code changes, only reformat.

Based on discussion #25

PR #45, opened 2026-04-23
Add .editorconfig to help enforce tab indentation

sbeitzel:
I was reading the discussion in PR #4 and saw that the maintainer prefers tabs over spaces for indentation. I thought, "Gee, if the project had an [.editorconfig](https://editorconfig.org/#pre-installed) then contributors might be less likely to cause this particular headache." So here I am, offering a starting .editorconfig.


Drafteed (changes requested):
This will conflict with #42

anszom:
The formatting was quite a mess so at this point I've given up and gladly acepted the prettier solution (#42). I can add the `.editorconfig` if you adapt it to match the current `.prettierrc` settings.

PR #51, opened 2026-05-09
RAC_056905_WW "climate" current action and fix for missing values-detected features

maciejsszmigiero:
Some of the features in `RAC_056905_WW` device do not have known capability bits.
This is true for example for device error code, EEV opening, nominal capacity and the `IDUThermoOnOff` flag described below.

The only known way to detect them is to check for the presence of the relevant tag in the device values response before creating and setting HA configuration of the device.

For this, send an initial values query at device initialization time once capabilities are received and define a `valuesReceived()` method which the device model child class can use to act on them.

This fixes missing presence of values-detected `RAC_056905_WW` features.

This wasn't noticed when preparing #36 since the development version created the relevant HA components unconditionally. This meant that when auto-detection of these features was added the fact that it never triggers wasn't apparent since the HA components were already there.

----

HA `climate` component has a notion of "current action". This is supposed to show what the device is currently actually doing, rather than what it has been set to do.

For example, the device could be set to cooling but is not actively cooling at this moment as the room temperature is already below the temperature set point.

Some IDU models (like the _Deluxe_ class for example) support `IDUThermoOnOff` flag.
This flag gets reported as `0` if the device is not actively performing the requested action due to its thermostat reporting the room as already too cold/warm.
Other IDU models (like the _Standard2_ class for example) have an (unnamed) value at tag `0x6c` that seems to work the same way.

This PR adds current action support to `RAC_056905_WW` HA `climate` component - even for IDUs that currently do not support that thermostat flag since it still results in nicer HA charts and `climate` widget dashboard appearance.

Edit: Also support the _Standard2_ class tag `0x6c` in the "current action" component.

### DIFF
--- a/rethink/cloud/devices/RAC_056905_WW.ts
+++ b/rethink/cloud/devices/RAC_056905_WW.ts
@@ -169,6 +169,14 @@ export default class Device extends TLVDevice {
         this.sendFilterQuery()
     }
 
+    getPowerTLV() {
+        return this.raw_clip_state[0x1f7]
+    }
+
+    getModeTLV() {
+        return this.raw_clip_state[0x1f9]
+    }
+
     initMakeSetConfig() {
         const config: DeviceDiscovery & { components: { climate: ClimateComponent } } = allowExtendedType({
             ...HADevice.config(this.meta, { name: 'LG Air Conditioner' }),
@@ -224,7 +232,7 @@ export default class Device extends TLVDevice {
             comp: 'climate',
             read_xform: (raw) => {
                 const modes2ha = ['cool', 'dry', 'fan_only', undefined, 'heat', undefined, 'auto']
-                if (this.raw_clip_state[0x1f7] === 0) return 'off'
+                if (this.getPowerTLV() === 0) return 'off'
                 return modes2ha[raw]
             },
             read_callback: (val) => {
@@ -585,22 +593,22 @@ export default class Device extends TLVDevice {
                 if (!this.jetMode) return 0
 
                 /* ON */
-                if (jetCool && this.raw_clip_state[0x1f9] === 0) return 1
-                if (jetHeat && this.raw_clip_state[0x1f9] === 4) return 2
+                if (jetCool && this.getModeTLV() === 0) return 1
+                if (jetHeat && this.getModeTLV() === 4) return 2
                 return 0
             },
             read_xform: (raw) => {
-                if (jetCool && this.raw_clip_state[0x1f9] === 0 && raw == 1) return 'ON'
-                if (jetHeat && this.raw_clip_state[0x1f9] === 4 && raw == 2) return 'ON'
+                if (jetCool && this.getModeTLV() === 0 && raw == 1) return 'ON'
+                if (jetHeat && this.getModeTLV() === 4 && raw == 2) return 'ON'
                 return 'OFF'
             },
             read_callback: (val) => {
                 // Ignore read value if not running
-                if (this.raw_clip_state[0x1f7] === 0 || this.raw_clip_state[0x1f7] == null) return false
+                const powerTLV = this.getPowerTLV()
+                if (powerTLV === 0 || powerTLV == null) return false
 
                 // Ignore read value if not in the right mode
-                if (!((jetCool && this.raw_clip_state[0x1f9] === 0) || (jetHeat && this.raw_clip_state[0x1f9] === 4)))
-                    return false
+                if (!((jetCool && this.getModeTLV() === 0) || (jetHeat && this.getModeTLV() === 4))) return false
 
                 this.jetMode = val === 'ON'
                 return true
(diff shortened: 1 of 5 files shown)


rethink/cloud/devices/RAC_056905_WW.ts
```diff
@@ -81,7 +82,15 @@ export default class Device extends TLVDevice {
         return tlvArray.some(({ t, v }) => t === 0x2da)
     }
 
-    capabilityReceived() {
+    isValuesResponse(tlvArray: TLV.TLV[]) {
+        /* power */
+        return tlvArray.some(({ t, v }) => t === 0x1f7)
```
anszom: This might hit some false positives in the odd case where the device is power-switched simultaneously with a rethink query. Maybe add a check for some other values appearing together with 1f7, or add a simple && tlvArray.length > threshold check?

rethink/cloud/devices/RAC_056905_WW.ts
```diff
@@ -7,14 +7,14 @@ import * as TLV from '@/util/tlv'
...
+type PowerModeChangeHook = () => void
 type CheckMode = (arg: number) => boolean
 export default class Device extends TLVDevice {
     meta: Metadata
     initialValuesReceived: boolean = false
-    powerOnHooks: PowerOnHook[] = []
-    powerStatePrev?: boolean
-    modeChangeHooks: PowerOnHook[] = []
+    powerChangeHooks: PowerModeChangeHook[] = []
+    powerStatePrev?: string
```
anszom: Can this be anything other than ON / OFF ? Why change to string then?
If other values are required as well, use a type such as 'ON' | 'OFF' | 'Foo' | 'Bar'

anszom (changes requested):
Looks good, just a few details to fix.

rethink/cloud/devices/RAC_056905_WW.ts
maciejsszmigiero: > Why change to string then?

I wanted to avoid parsing the received string `val` into a boolean in `read_callback` by just storing it as-is.
But I have changed `powerStatePrev` back to boolean now.

rethink/cloud/devices/RAC_056905_WW.ts
maciejsszmigiero: That's a very good point.

I added a reply length condition of `>= 10` here since real power switch notifications seem to have 2-3 tags, while a real values query result is 40+ tags long - so there's quite a safety margin here.

rethink/cloud/devices/RAC_056905_WW.ts
```diff
@@ -160,6 +169,51 @@ export default class Device extends TLVDevice {
...
+        let action: string | undefined = undefined
+        if (this.getPowerTLV() === 0) {
+            action = 'off'
+        } else if ((modeTLV === 0 || modeTLV === 1 || modeTLV === 4 || modeTLV === 6) && !iduRunning) {
+            action = 'idle'
+        } else if (modeTLV === 6) {
+            // TODO: figure out how to detect the actual running mode in Auto
+            // For now, clear the reported action.
+            action = 'None'
+        } else {
```
Drafteed, line 191: Is there any way to get the status of the four-way valve from the units? The most reasonable approach would be to use it to determine which mode the air conditioner is currently operating in.

If that's not possible, it might be possible to infer the status based on the target and current temperatures, although that probably wouldn't be very reliable.

Drafteed (approved):
LGTM 👍
Tested on LG TC12GQ - everything is fine.

rethink/cloud/devices/RAC_056905_WW.ts
maciejsszmigiero: > Is there any way to get the status of the four-way valve from the units?

Unfortunately, I don't know of any way to get the 4way valve status via ThinQ interface. Would still not allow differentiating between _cooling_ and _drying_ actual modes selected by the _Auto_ mode though.

> it might be possible to infer the status based on the target and current temperatures, although that probably wouldn't be very reliable.

Yeah, especially that these IDUs do fuzzy turn-on/turn-off temperature thresholds.

Hopefully at some point the proper tag to tell with high accuracy the actual IDU operation mode will be found.

Drafteed: @maciejsszmigiero, I ran into a similar issue in a comparable component for Haier AC units. I was able to resolve it by using the indoor coil temperature: if it is higher than the current room temperature, the unit is heating; otherwise, it is cooling. It may not be a perfect solution, but it works nice.

Check this out: https://github.com/paveldn/haier-esphome/pull/116/changes#diff-75d0fb113093e83ebdd53d1975d6623b8b58042ac44fc2348d1d60da0f1b63edR1003-R1014

It may be possible to implement something similar here, especially since the pipe temps is available.

maciejsszmigiero: I'm testing some running mode autodetection heuristics now, we'll see how it goes.

PR #49, opened 2026-05-03
Add `F_V__F___W.B_1QEUK` support & refactor washers

Drafteed:
## Summary

This PR adds support for the `F_V__F___W.B_1QEUK` LG washer-dryer and aligns washer Home Assistant entities across existing ThinQ1/ThinQ2 device implementations.

## Changes

- Added shared washer mappings in `washer_common.ts` for errors, states, courses, temperatures, spin speeds, and drying modes;
- Reused the shared washer metadata across ThinQ1/ThinQ2 washers: `WTDN3`, `F_V8_Y___W.B_2QEUK`, and `Y_V8_Y___W.B32QEUK`;
- Standardized washer HA entities:
  - Power Switch
  - Start/Pause buttons
  - Status and error diagnostics
  - Course, Temperature, Spin speed and Cycle count
  - Remote start, Door lock, Initial/Remaining time
  - Energy (where supported)
- Added `F_V__F___W.B_1QEUK` ThinQ2  washer-dryer support.

## Note

This PR contains some soft breaking changes. You may need to re-add the MQTT device in HA to discard duplicate/dead entitites.

@ToniH1987 @moutasem1989 please check this. 

If you are using Docker, you can do this via my fork:
```
docker run -d --name rethink --restart unless-stopped --network host -v $(pwd)/rethink-docker-config:/app/data ghcr.io/drafteed/rethink:dev
```

or `docker-compose.yml`:
```
services:
  rethink:
    container_name: rethink
    image: ghcr.io/drafteed/rethink:dev
    restart: unless-stopped
    volumes:
      - ./data:/app/data
    network_mode: host
```

PR #52, opened 2026-05-10
Add Washer&Dryer (CV74J7S2QA) support

Jonesus:
I have a LG ThinQ Washer&Dryer (10,5kg/7kg) (model no. CV74J7S2QA) that wasn't fully supported before. I saw that it worked well enough by monkey-patching the supported device list to also allow `F_VB_F___W.B_2QEUK`, so I got inspired to make it work better. The model existing in current codebase lacked drying capabilities which I wanted to see through HA myself.

Disclaimer; the code is slopped together with Claude, but I reviewed it and this PR's text (and any communication in general) is hand-written. I couldn't find a CONTRIBUTING.md or similar that would've forbidden it :)

### DIFF
--- a/rethink/cloud/devices/F_VB_F___W.B_2QEUK.ts
+++ b/rethink/cloud/devices/F_VB_F___W.B_2QEUK.ts
@@ -1,0 +1,284 @@
+import HADevice from './base'
+import { Device as Thinq2Device } from '../thinq2/device'
+import { type Connection } from '../homeassistant'
+import { type Metadata } from '../thinq'
+import { allowExtendedType } from '@/util/casting'
+import AABBDevice from './aabb_device'
+import { ERRORS, STATES, COURSES, TEMPERATURES, SPINS, DRYING_MODES, DOSES } from './washer_common'
+
+export default class Device extends AABBDevice {
+    constructor(HA: Connection, thinq: Thinq2Device, meta: Metadata) {
+        super(HA, thinq)
+        this.setConfig(
+            allowExtendedType({
+                ...HADevice.config(meta, { name: 'LG Washer' }),
+                components: {
+                    power: {
+                        platform: 'switch',
+                        unique_id: '$deviceid-power',
+                        state_topic: '$this/power',
+                        command_topic: '$this/power/set',
+                        name: '',
+                        icon: 'mdi:washing-machine',
+                    },
+                    start: {
+                        platform: 'button',
+                        unique_id: '$deviceid-start',
+                        command_topic: '$this/start/set',
+                        payload_press: '',
+                        name: 'Start',
+                        icon: 'mdi:play-circle-outline',
+                    },
+                    pause: {
+                        platform: 'button',
+                        unique_id: '$deviceid-pause',
+                        command_topic: '$this/pause/set',
+                        payload_press: '',
+                        name: 'Pause',
+                        icon: 'mdi:pause-circle-outline',
+                    },
+                    status: {
+                        platform: 'sensor',
+                        unique_id: '$deviceid-status',
+                        state_topic: '$this/status',
+                        name: 'Status',
+                        icon: 'mdi:state-machine',
+                        device_class: 'enum',
+                        options: STATES.filter((a) => a !== undefined),
+                    },
+                    error: {
+                        platform: 'binary_sensor',
+                        unique_id: '$deviceid-error',
+                        state_topic: '$this/error',
+                        name: 'Error',
+                        icon: 'mdi:check-circle',
+                        device_class: 'problem',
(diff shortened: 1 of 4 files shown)


Jonesus:
I now realized that I should've checked the existing PRs :sweat_smile: There's for sure overlap with #49 , so if you want to get that merged first we can go with that. I'm also interested in continuing work to make the entities controllable via Home Assistant, but I'll wait on that for now to avoid merge conflict hassles.

anszom:
That PR was just merged. About the washer functions I think it's best if you would coordinate with @Drafteed & his work. I don't have a Thinq-compatible washer :P

Drafteed (commented):
Hi! Thanks for your contribution. 

Two things:

First, we need to pull in the latest changes and use the new shared enums, while keeping the sensors style consistent with the other models.

Second thing: I'm not sure we should add sensors related to drying / detergent dosing / etc. to `F_VB_F___W.B_2QEUK`, since this model may not support those features (maybe @ToniH1987 can tell more). So I would create a new device specifically for `F_V8_Y___W.B_2QEUK`.

Jonesus:
Thanks for the feedback!

> First, we need to pull in the latest changes and use the new shared enums, while keeping the sensors style consistent with the other models.

Sure thing :+1: 

> Second thing: I'm not sure we should add sensors related to drying / detergent dosing / etc. to F_VB_F___W.B_2QEUK, since this model may not support those features (maybe @ToniH1987 can tell more). So I would create a new device specifically for F_V8_Y___W.B_2QEUK.

Fair point, let's go with that unless we hear otherwise from Toni.

I'll get the changes done later this week!

Jonesus:
Sorry for the delay, time flies. The PR is now rebased and there's new device for the washer/dryer, now mostly copied from `F_V8_Y___W.B_2QEUK.ts` with the drier-functionality added, tested with my physical device to be working, although with the shared enums there's a few programs where the physical sticker doesn't match the text exposed to HA, but that's something that probably varies too much between units to be feasibly fixed :)

rethink/cloud/devices/F_VB_F___W.B_2QEUK.ts
```diff
@@ -0,0 +1,286 @@
+import HADevice from './base'
+import { Device as Thinq2Device } from '../thinq2/device'
+import { type Connection } from '../homeassistant'
+import { type Metadata } from '../thinq'
+import { allowExtendedType } from '@/util/casting'
+import AABBDevice from './aabb_device'
+import { ERRORS, STATES, COURSES, TEMPERATURES, SPINS, DRYING_MODES } from './washer_common'
+
+const DOSES = ['Off', 'Low', 'Medium', 'High']
```
Drafteed: Let's move `DOSES` to `washer_common.ts`.

Drafteed (approved):
Looks good 👍

rethink/cloud/devices/F_VB_F___W.B_2QEUK.ts
```diff
@@ -0,0 +1,284 @@
...
+import { allowExtendedType } from '@/util/casting'
+import AABBDevice from './aabb_device'
+import { ERRORS, STATES, COURSES, TEMPERATURES, SPINS, DRYING_MODES, DOSES } from './washer_common'
+
+export default class Device extends AABBDevice {
+    constructor(HA: Connection, thinq: Thinq2Device, meta: Metadata) {
+        super(HA, thinq)
+        this.setConfig(
+            allowExtendedType({
+                ...HADevice.config(meta, { name: 'LG Washer/Dryer' }),
```
Drafteed: Please fix the formatting and one more thing:
```suggestion
                ...HADevice.config(meta, { name: 'LG Washer' }),
```

Drafteed (changes requested):
Please also add a test similar to other washing machines, including real samples from your device.

xBelladonna:
Please forgive the slightly off-topic post, but I came across this PR while trying to implement support for our `WVC5-1409W` washer/dryer (`F_V7_F___W.B__QEUK`). I noticed that the door lock has inverted logic, like yours, but that the child lock also followed this inverted logic. Your code appears to leave the child lock as normal; is this intentional? Our model appeared compatible enough that I'd planned to re-use your class, but if there are discrepancies like this, I'll have to make a new device.

Jonesus:
@xBelladonna yeah, unfortunately the child lock works "normally" for my machine :sweat_smile:

Jonesus:
@Drafteed all right, got the formatter run and tests with the intercepted messages

Drafteed (approved):
LGTM 👍

anszom:
thanks!

PR #60, opened 2026-05-30
RAC_056905_WW improvements vol. 3 including pipe and ODU air temperatures

maciejsszmigiero:
This PR contains volume 3 of improvements to the RAC_056905_WW device support.

The most important improvement is the addition of pipe and ODU air temperatures.

Specifically, it adds IDU liquid pipe and gas pipe temperatures, ODU heat exchanger (`HEX` as LGMV calls it) temperature and ODU air temperature.
The exact value -> temperature mapping matches what LGMV displays. 

For this reporting to work decently a variable polling interval is needed since many IDUs do not send timely notifications about changes to these or even do not notify about them at all.

The default polling interval of 15 minutes is almost useless for measuring pipe and ODU air temperatures as these change rapidly when the compressor is running.
That's why, for example LGMV always polls these values roughly every second.

This PR is conservative and polls every a bit less than half a minute when IDU is actively running as this matches well the observed ODU parameter recalculation intervals: target compressor Hz seems to be recalculated every 30 seconds, while target EEV openings seems to be recalculated every 30 seconds during transient periods.

When IDU stops actively running the polling interval is reset to the default 15 minute one after 15 minutes as things should have had returned to steady idle state by this time.

---

There are also a few minor changes in this PR, mostly relating to various text descriptions and a trivial increase of max sleep timer setting to 15 hours which helps with running with extra-low fan speed for longer since for many IDUs setting the sleep timer is the only way to get this fan speed.

---

Please refer to individual commit messages for details.


rethink/cloud/devices/RAC_056905_WW.ts
```diff
@@ -717,14 +785,20 @@ export default class Device extends TLVDevice {
...
     addOptionalSensorField(
         config: DeviceDiscovery,
-        id: number,
+        ids: number | number[],
         name: string,
         desc: string,
         icon?: string,
         extra?: Record<string, unknown>,
         read_xform?: FieldDefinition['read_xform'],
+        reject_zero: boolean = false,
```
anszom: I think it would be better to fold `reject_zero` into the read_xform.

rethink/cloud/devices/RAC_056905_WW.ts
```diff
@@ -717,14 +785,20 @@ export default class Device extends TLVDevice {
...
     ) {
-        if (this.raw_clip_state[id] == null) return
+        if (typeof ids === 'number') {
+            ids = [ids]
+        }
+
+        let id = ids.find((val) => this.raw_clip_state[val] != null && (!reject_zero || this.raw_clip_state[val] !== 0))
```
anszom: ... here you could check read_xform (...) !== undefined, instead of testing reject_zero

rethink/cloud/devices/RAC_056905_WW.ts
```diff
@@ -429,6 +431,72 @@ export default class Device extends TLVDevice {
...
+            'pipeintemp',
+            'Pipe liquid temperature',
+            'mdi:pipe',
+            {
+                device_class: 'temperature',
+                unit_of_measurement: '°C',
+                suggested_display_precision: 2,
+            },
+            (raw) => racPipeTemp[255 - raw],
+            true /* -100 C is not plausible */,
```
anszom: maybe it would be simpler to drop the `-100` entry. Then racPipeTemp[255-raw] would implicitly return undefined for a zero value. See comment for `addOptionalSensorField`

anszom (changes requested):

rethink/cloud/devices/RAC_056905_WW.ts
maciejsszmigiero: Made providing `read_xform` work like providing `reject_zero = true` and dropped explicit `reject_zero`.

Added a comment about such side effect near `read_xform` parameter.

rethink/cloud/devices/RAC_056905_WW.ts
maciejsszmigiero: The condition now checks for `read_xform` being provided instead of `reject_zero` being `true`.

rethink/cloud/devices/RAC_056905_WW.ts
maciejsszmigiero: Now that `reject_zero` has been folded into `read_xform` I would prefer keeping the last `racPipeTemp` entry since in principle it is still a valid general mapping value and  that `RAC_056905_WW` returns 0 raw value as _unknown_ looks more of a firmware/protocol implementation detail for that particular device type.

anszom: does a value of 255 really represent -100 degrees celsius? It doesn't look that way from reading the tables - the -100 value is discontinuous and appears to be a "sentinel" value. `undefined` would be a better representation for that, and would filter out cleanly at `read_xform`

rethink/cloud/devices/RAC_056905_WW.ts
anszom: Sorry, I didn't explain this clearly the first time.

read_xform already has the capability to filter out values by returning `undefined`.
Instead of special-casing zero, you need to simply pass a function which returns `undefined` for invalid values. As it happens:
```
(raw) => racAirTemp[255 - raw]
``` 

would be just the right expression for read_xform, if the table `racAirTemp` was one entry shorter. If you insist on filtering explicitly, it would be `(raw) => raw ? racAirTemp[255-raw] : undefined`

For the ids condition, you would just need to run the read_xform instead of testing for zero explicitly:
```
let id = ids.find(
            (val) => this.raw_clip_state[val] != null && (read_xform == null || read_xform(this.raw_clip_state[val]) !== undefined),
        )
```

rethink/cloud/devices/RAC_056905_WW.ts
maciejsszmigiero: You are right that the -100 C value at index 255 looks discontinuous, so I converted it to a sentinel instead.

rethink/cloud/devices/RAC_056905_WW.ts
maciejsszmigiero: Made use of the fact that index 255 of both mapping tables is undefined now, removed the `read_xform` wrapper in `addOptionalSensorField()` and made this function test the initial tag value against `read_xform(this.raw_clip_state[val])`.

Modified capacity sensor raw mapping accordingly to return `undefined` when the raw value is 0.

anszom:
looks good. Thanks!

PR #64, opened 2026-06-02
Add support for LG dehumidifier DHUM_056905_WW

BluSyn:
MQTT discovery and TLV control for the 056905 clip platform (0xA7/0x87).

Supported features:
- Power and operating modes: smart, jet, silent, spot, laundry
- Current and target humidity (30–70%)
- Fan speed (low/high) via select
- Ionizer, UVnano, and bucket light switches (not available on official integration)
- Off timer: off or 1–8 minutes

Includes capability-packet handling, unit tests from panel captures, and bridge registration.

Note: I can only be certain of compatibility with this exact model of dehumidifier: MD19GQGE0 (other models in this family have slightly different capabilities and controls exposed).

### DIFF
--- a/cloud/devices/DHUM_056905_WW.ts
+++ b/cloud/devices/DHUM_056905_WW.ts
@@ -256,7 +256,7 @@ export default class Device extends TLVDevice {
             write_xform: (val) => (val === 'ON' ? 1 : 0),
         })
 
-        this.addTimerField(config, 0x21b, 'off_timer', 'Sleep timer', 'mdi:bed-clock', 8)
+        this.addTimerField(config, 0x21b, 'off_timer', 'Sleep timer', 'mdi:bed-clock', 9)
 
         // Wire bare state_topic/command_topic (expected by humidifier platform) to our 'power' property
         const hum = (config.components as any).humidifier
@@ -274,7 +274,7 @@ export default class Device extends TLVDevice {
             name: desc,
             icon: icon,
             device_class: 'duration',
-            unit_of_measurement: 'min',
+            unit_of_measurement: 'h',
             min: 0,
             max: max,
             step: step,
@@ -283,17 +283,14 @@ export default class Device extends TLVDevice {
         config['components'][name] = comp
 
         /*
-         * Upon setting this field the device starts counting down and
-         * sends the remaining time (seconds); ceil to whole minutes for HA.
+         * Setpoint is hours×60 (minutes) in tlv, same as RAC timers on 0x21b.
+         * While counting down, notifies send remaining time in seconds.
          */
         this.addField(config, {
             id: id,
             name: '',
             comp: name,
-            read_xform: (raw) => {
-                const val = Math.ceil(raw / 60 / step) * step
-                return val > max ? 0 : val
-            },
+            read_xform: (raw) => Math.ceil(raw / 60 / step) * step,
             write_xform: (val) => Math.round(Number(val) * 60),
         })
     }
--- a/cloud/devices/DHUM_056905_WW.ts
+++ b/cloud/devices/DHUM_056905_WW.ts
@@ -93,13 +93,6 @@ export default class Device extends TLVDevice {
                     icon: 'mdi:fan',
                     options: ['low', 'high'],
                 },
-                off_timer: {
-                    platform: 'select',
-                    unique_id: '$deviceid-off_timer',
-                    name: 'Off timer',
-                    icon: 'mdi:timer-off',
-                    options: ['off', '1', '2', '3', '4', '5', '6', '7', '8'],
-                },
                 current_humidity: {
                     platform: 'sensor',
                     unique_id: '$deviceid-current_humidity',
@@ -261,23 +254,7 @@ export default class Device extends TLVDevice {
             write_xform: (val) => (val === 'ON' ? 1 : 0),
         })
 
-        // off timer (0x21b: 0=off; setpoint is minutes×60s; notifies may show seconds remaining e.g. 59→1 min)
-        this.addField(config, {
-            id: 0x21b,
-            name: '',
-            comp: 'off_timer',
-            read_xform: (raw) => {
-                if (!raw) return 'off'
-                const minutes = Math.ceil(raw / 60)
-                return minutes >= 1 && minutes <= 8 ? String(minutes) : 'off'
-            },
-            write_xform: (val) => {
-                if (val === 'off') return 0
-                const minutes = Number(val)
-                if (!Number.isFinite(minutes) || minutes < 1 || minutes > 8) return 0
-                return minutes * 60
-            },
-        })
+        this.addTimerField(config, 0x21b, 'off_timer', 'Sleep timer', 'mdi:bed-clock', 8)
 
         // Wire bare state_topic/command_topic (expected by humidifier platform) to our 'power' property
         const hum = (config.components as any).humidifier
@@ -285,6 +262,38 @@ export default class Device extends TLVDevice {
         hum.command_topic = '$this/humidifier-power/set'
 
         this.setConfig(config)
+    }
+
+    addTimerField(config: DeviceDiscovery, id: number, name: string, desc: string, icon: string, max: number) {
+        const step = 1
+        const comp = {
+            platform: 'number',
+            unique_id: '$deviceid-' + name,
+            name: desc,
+            icon: icon,
+            device_class: 'duration',
+            unit_of_measurement: 'min',
+            min: 0,
+            max: max,
+            step: step,
+            mode: 'slider',
+        } as const
+        config['components'][name] = comp
+
+        /*
+         * Upon setting this field the device starts counting down and
+         * sends the remaining time (seconds); ceil to whole minutes for HA.
+         */
+        this.addField(config, {
+            id: id,
+            name: '',
+            comp: name,
+            read_xform: (raw) => {
+                const val = Math.ceil(raw / 60 / step) * step
+                return val > max ? 0 : val
+            },
+            write_xform: (val) => Math.round(Number(val) * 60),
+        })
     }
 
     private fanSpeedFromClip(raw?: number): string {
--- a/cloud/devices/DHUM_056905_WW.ts
+++ b/cloud/devices/DHUM_056905_WW.ts
@@ -100,6 +100,7 @@ export default class Device extends TLVDevice {
                     device_class: 'humidity',
                     unit_of_measurement: '%',
                     state_class: 'measurement',
+                    state_topic: '$this/humidifier-current_humidity',
                 },
                 bucket_full: {
                     platform: 'binary_sensor',
@@ -199,8 +200,9 @@ export default class Device extends TLVDevice {
         // current humidity (observed on 0x1fd in device state packets, e.g. 0x30=48%)
         this.addField(config, {
             id: 0x1fd,
-            name: '',
-            comp: 'current_humidity',
+            name: 'current_humidity',
+            comp: 'humidifier',
+            state_topic: 'topic',
             writable: false,
         })
 
--- a/cloud/devices/DHUM_056905_WW.ts
+++ b/cloud/devices/DHUM_056905_WW.ts
@@ -9,6 +9,9 @@ import HADevice from './base'
 /** TLV tags present in capability (0xA7/0x01) packets — store state but do not publish as entity values. */
 const CAPS_ONLY_TAGS = new Set([0x2d5, 0x2d6, 0x336, 0x2e5, 0x2e6, 0x2da])
 
+/** Observed on bucket-empty notify when the tank is reinstalled (0x2b1=256, 0x2b2=0). */
+const BUCKET_EMPTIED_EVENT = 256
+
 /** Entering these modes resets fan speed to low (high remains user-selectable). */
 const SILENT_MODES = new Set([2, 19])
 
@@ -47,6 +50,8 @@ export default class Device extends TLVDevice {
     modePrev?: string
     modeClipPrev?: number
     initialValuesReceived = false
+    /** Last bucket-full state published to HA (retained). */
+    bucketFullHaState?: boolean
 
     constructor(HA: Connection, thinq: Thinq2Device, meta: Metadata) {
         super(HA, thinq)
@@ -102,6 +107,16 @@ export default class Device extends TLVDevice {
                     device_class: 'humidity',
                     unit_of_measurement: '%',
                     state_class: 'measurement',
+                },
+                bucket_full: {
+                    platform: 'binary_sensor',
+                    unique_id: '$deviceid-bucket_full',
+                    name: 'Bucket full',
+                    icon: 'mdi:water-alert',
+                    device_class: 'problem',
+                    payload_on: 'ON',
+                    payload_off: 'OFF',
+                    state_topic: '$this/bucket_full-',
                 },
             },
         })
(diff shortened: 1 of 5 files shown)


rethink/cloud/devices/DHUM_056905_WW.ts
```diff
@@ -0,0 +1,352 @@
...
+                    return undefined
+                } else {
+                    this.setProperty('humidifier-power', 'ON')
+                }
+                const mode = normalizeHaMode(val)
+                const clip = HA_TO_CLIP_MODE[mode] ?? Number(val)
+                if (mode === 'Silent' && (this.modeClipPrev == null || !SILENT_MODES.has(this.modeClipPrev))) {
+                    this.raw_clip_state[0x1fa] = 2
+                    this.publishFanSpeedState('low')
+                }
```
anszom, line 206: help me understand this. The dehumidifier, when switched into mode 2 or 9, changes the fan speed to 'low', but doesn't announce the change?

rethink/cloud/devices/DHUM_056905_WW.ts
```diff
@@ -0,0 +1,352 @@
...
+                this.publishFanSpeedState(typeof val === 'string' ? val : String(val))
+                return false
+            },
+            write_xform: (val) => {
+                const modes2clip: Record<string, number> = { low: 2, high: 6 }
+                return modes2clip[val] ?? Number(val)
+            },
+            write_callback: (val) => {
+                if (val !== 2 && val !== 6) return false
+                this.sendFanSpeedTlvs(val)
```
anszom, line 231: I think this could be handled by the `write_attach` option.

anszom (commented):
Thanks for the PR! See a couple of comments below

rethink/cloud/devices/DHUM_056905_WW.ts
BluSyn: Yes, sort-of. It defaults to low on Silent mode, and does eventually announce fan speed change on next state update, but in my testing there was a delay from a few seconds to tens of seconds to reflect fan speed change in homeassistant UI. By mimicking that behavior here there's no delay. It's not strictly necessary, but I found it a better UX to make this explicit.

rethink/cloud/devices/DHUM_056905_WW.ts
BluSyn: Will test this

rethink/cloud/devices/DHUM_056905_WW.ts
```diff
@@ -0,0 +1,352 @@
...
+import * as TLV from '@/util/tlv'
+import HADevice from './base'
+
+/** TLV tags present in capability (0xA7/0x01) packets — store state but do not publish as entity values. */
+const CAPS_ONLY_TAGS = new Set([0x2d5, 0x2d6, 0x336, 0x2e5, 0x2e6, 0x2da])
+
+/** Entering these modes resets fan speed to low (high remains user-selectable). */
+const SILENT_MODES = new Set([2, 19])
+
+const HA_MODES = ['Smart', 'Jet', 'Silent', 'Spot', 'Laundry'] as const
```
Drafteed, line 22: Consider to (partial) match HA modes:
https://developers.home-assistant.io/docs/core/entity/humidifier/#modes

e.g.:
Smart -> Auto
Jet -> Boost
Silent -> Sleep / Eco

rethink/cloud/devices/DHUM_056905_WW.ts
```diff
@@ -0,0 +1,352 @@
...
+                    options: ['off', '1', '2', '3', '4', '5', '6', '7', '8'],
+                },
+                current_humidity: {
+                    platform: 'sensor',
+                    unique_id: '$deviceid-current_humidity',
+                    name: 'Current humidity',
+                    device_class: 'humidity',
+                    unit_of_measurement: '%',
+                    state_class: 'measurement',
+                },
```
Drafteed, line 148: Why is the current humidity published as a separate sensor instead of being exposed through the `humidifier` entity?

rethink/cloud/devices/DHUM_056905_WW.ts
```diff
@@ -0,0 +1,352 @@
...
+                    icon: 'mdi:fan',
+                    options: ['low', 'high'],
+                },
+                off_timer: {
+                    platform: 'select',
+                    unique_id: '$deviceid-off_timer',
+                    name: 'Off timer',
+                    icon: 'mdi:timer-off',
+                    options: ['off', '1', '2', '3', '4', '5', '6', '7', '8'],
+                },
```
Drafteed: Maybe this should have been implemented as a number entity with limits and `unit_of_measurement: 'h'`, as  how it is done for AC?https://github.com/anszom/rethink/blob/f7f7ae49abe23478f432c9718813e76f5e0428c2/rethink/cloud/devices/RAC_056905_WW.ts#L610-L623

rethink/cloud/devices/DHUM_056905_WW.ts
BluSyn: Better if options in HASS match options on device control? Generic names can make for confusing UI.

rethink/cloud/devices/DHUM_056905_WW.ts
BluSyn: Should be both. Had that working before. Will fix.

rethink/cloud/devices/DHUM_056905_WW.ts
BluSyn: Good call. Wasn't sure the best implementation for this, but wanted to expose somehow since it was available.

rethink/cloud/devices/DHUM_056905_WW.ts
BluSyn: `write_attach` supports one TLV per tag; fan speed needs repeated 0x2d7/0x2d8/0x2d9 per mode. Was tricky to get this to work. Could add a new primitive to `FieldDefinition` in `tlv_device.ts` for this use-case, but tried to avoid any interface changes.

anszom: Why one? `write_attach` is an array

anszom: Ok, I *think* I get it now, some 'T' values need to be sent out multiple times to achieve the desired result...

rethink/cloud/devices/DHUM_056905_WW.ts
```diff
@@ -0,0 +1,397 @@
...
+                      [17, 6],
+                      [18, 6],
+                      [20, 6],
+                      [21, 6],
+                      [22, 6],
+                  ]
+        const tlvs = [{ t: 0x1fa, v: fan }, ...modes.flatMap(([m, f]) => modeFan(m, f))]
+        for (const { t, v } of tlvs) this.raw_clip_state[t] = v
+        return tlvs
+    }
```
anszom, line 370: It is quite difficult to understand what is happening here. 

From what I can decipher, it's something like
```
modeFan(17, fan),
modeFan(18, fan), 
modeFan(20, fan), 
modeFan(21, 6), // always 6 - why? 
modeFan(22, fan)
```

since it's quite obvious that 2d7 2d8 and 2d9 don't represent any gobally understood state, does it make sense to store them in `raw_clip_state` ?

Please rewrite this function so that the meaning would be evident. Also, since this feature is one of more complex parts of this implementation, please add an unit test to cover this.

3735943886:
I have this device too and this PR covers nearly everything I need, so I'd like to see it land.
I've mapped the tags on my unit against the LG cloud's decoded property names (the ThinQ
monitoring feed labels each wire tag with its canonical `airState.*` name), and that answers
some of the open questions here.

### 0x2d7 / 0x2d8 / 0x2d9 are the supported-mode capability table

> some 'T' values need to be sent out multiple times to achieve the desired result...
> `modeFan(21, 6), // always 6 - why?`
> since it's quite obvious that 2d7 2d8 and 2d9 don't represent any globally understood
> state, does it make sense to store them in `raw_clip_state`?

They're a repeating 3-tuple table the device uses to declare which modes it supports and what
fan speed each one runs at — which is why the same tags repeat:

- `0x2d7` = mode id (17 Smart, 18 Jet, 19 Silent, 20 Spot, 21 Laundry, and 22, which appears in
  the table on my unit but isn't exposed in the app)
- `0x2d8` = 0 in every capture I have
- `0x2d9` = the fan speed that mode runs at (2 = low, 6 = high)

So `modeFan(21, 6)` is "always 6" because **Laundry mode only supports high fan** — it's a
property of that mode, not a special case. An explicit `MODE_FAN_CAPS` table named after that
should make the function self-explanatory without changing behaviour.

And you're right that this isn't device state: it's a capability declaration, so rather than
living in `raw_clip_state` it would be more natural to parse it into the `modes` list instead
of hardcoding it — that would also cover sibling models with a different mode set.

Related data point on the fan question: on my unit a fan-speed change *is* announced — the
notify frame starts with `0x1fa` = the new value, followed by the resent capability table
above.

### Current humidity looks like it's on `0x336`

On my unit the cloud labels `0x336` as **`airState.humidity.current`** — which lines up with
the note already in the code:

> `// 0x336 tracks humidity in live packets, not bucket level (was 45–50% during testing).`

45–50 % is the room humidity, but `processKeyValue` currently swallows `0x336` and publishes
`0x1fd` as `current_humidity` instead. On my unit `0x1fd` decodes to a different cloud property,
so it's probably worth checking which of the two actually tracks the humidity shown in the app.

BluSyn:
> I have this device too and this PR covers nearly everything I need, so I'd like to see it land. I've mapped the tags on my unit against the LG cloud's decoded property names (the ThinQ monitoring feed labels each wire tag with its canonical `airState.*` name), and that answers some of the open questions here.
> 
> ### 0x2d7 / 0x2d8 / 0x2d9 are the supported-mode capability table
> > some 'T' values need to be sent out multiple times to achieve the desired result...
> > `modeFan(21, 6), // always 6 - why?`
> > since it's quite obvious that 2d7 2d8 and 2d9 don't represent any globally understood
> > state, does it make sense to store them in `raw_clip_state`?
> 
> They're a repeating 3-tuple table the device uses to declare which modes it supports and what fan speed each one runs at — which is why the same tags repeat:
> 
> * `0x2d7` = mode id (17 Smart, 18 Jet, 19 Silent, 20 Spot, 21 Laundry, and 22, which appears in
>   the table on my unit but isn't exposed in the app)
> * `0x2d8` = 0 in every capture I have
> * `0x2d9` = the fan speed that mode runs at (2 = low, 6 = high)
> 
> So `modeFan(21, 6)` is "always 6" because **Laundry mode only supports high fan** — it's a property of that mode, not a special case. An explicit `MODE_FAN_CAPS` table named after that should make the function self-explanatory without changing behaviour.
> 
> And you're right that this isn't device state: it's a capability declaration, so rather than living in `raw_clip_state` it would be more natural to parse it into the `modes` list instead of hardcoding it — that would also cover sibling models with a different mode set.
> 
> Related data point on the fan question: on my unit a fan-speed change _is_ announced — the notify frame starts with `0x1fa` = the new value, followed by the resent capability table above.
> 
> ### Current humidity looks like it's on `0x336`
> On my unit the cloud labels `0x336` as **`airState.humidity.current`** — which lines up with the note already in the code:
> 
> > `// 0x336 tracks humidity in live packets, not bucket level (was 45–50% during testing).`
> 
> 45–50 % is the room humidity, but `processKeyValue` currently swallows `0x336` and publishes `0x1fd` as `current_humidity` instead. On my unit `0x1fd` decodes to a different cloud property, so it's probably worth checking which of the two actually tracks the humidity shown in the app.

These are great suggestions, thanks for deep dive. I keep forgetting to follow up on this PR, as once I got this working for my device I moved on to other homeassistant projects 🤷‍♂️ . 

I will try to follow up with suggested fixes soon so we can get this merged. (Also looks like branch needs to be rebased to fix conflicts)

BluSyn:
Rebased against latest master. Fully tested and working on my device.

I refactored the state logic per suggestions, and added some more tests to cover different mode cases. Also cleaned up some other confusing logic.

Regarding 0x336 vs 0x1fd as current_humidity: I switched to 0x336 for this field, and I get the same output as before. Not sure why this is the case, or if my humidity and temp values are closely matched. (Its not clear 0x1fd *is* temp, but it is on other platforms, so I'm assuming thats the confusion..)

@3735943886 review on your device if you have additional feedback.

Otherwise I think this is good to merge @anszom

3735943886:
This looks good to merge.** One thing I'd ask for before it does — renaming the `ionizer` switch.

### 0x360 is airRemoval, not the ioniser — the one thing worth doing before merge

1. The model JSON names it that (`airRemoval_state_tlv_864`, 864 = 0x360).
2. The ioniser on this model is an operation **mode**, not a switch: `airState.opMode` wire 22
   is `@AP_MAIN_MID_OPMODE_IONIZER_W`.

So as it stands there are two different things called the ioniser, and the switch isn't the
one the app means.

What I'm confident about is the cloud property name and that it isn't the ioniser; the
user-facing label is your call. I went with `sterilization` / "Air sterilization" locally to
match the air purifier profile, but if the dehumidifier's app calls this something else in
your locale, that name should win.

tests/cloud/devices/DHUM_056905_WW.test.ts
```diff
@@ -0,0 +1,330 @@
...
+
+const UV_ON_NOTIFY_HEX = '000004000000A702044B02A88184D7'
+const UV_OFF_NOTIFY_HEX = '000004000000A702044A08A8808C90388CD041E991'
+
+const BUCKET_LIGHT_ON_NOTIFY_HEX = '000004000000A702048C0287817086'
+const BUCKET_LIGHT_OFF_NOTIFY_HEX = '000004000000A702048A028780473E'
+
+// Sleep timer (0x21b) countdown notifies: remaining seconds in tlv
+const SLEEP_TIMER_59S_NOTIFY_HEX = '000004000000A70204ED0386D03BB028' // ~1 h displayed
+const SLEEP_TIMER_299S_NOTIFY_HEX = '000004000000A70204F30486E0012BC8DA' // ~5 h displayed
```
anszom: Something's off here. Either 59S = 1 min, or 59M = 1 hour.

cloud/homeassistant.ts
```diff
@@ -193,3 +193,12 @@ export type ClimateComponent = ComponentInfo & {
...
 }
+
+export type HumidifierComponent = ComponentInfo & {
+    platform: 'humidifier'
+    device_class?: 'dehumidifier' | 'humidifier'
+    modes?: string[]
+    min_humidity?: number
+    max_humidity?: number
+    current_humidity_topic?: string
+}
```
anszom, line 208: This isn't used anywhere?

cloud/devices/DHUM_056905_WW.ts
```diff
@@ -0,0 +1,426 @@
...
+                }
+                if (mode != null) this.modeClipPrev = mode
+                return true
+            },
+            write_xform: (val) => {
+                if (val === 'off' || val === undefined) {
+                    this.setProperty('humidifier-power', 'OFF')
+                    return undefined
+                } else {
+                    this.setProperty('humidifier-power', 'ON')
```
anszom: This will - I think - send unnecessary commands to the device, first with an old mode, then with a new one. Better to simply set `raw_clip_state[0x1f7]` here to set power on, it will get attached to the write (only for the power-on branch).

Still, this needs to  be validated on an actual device.

cloud/devices/DHUM_056905_WW.ts
```diff
@@ -0,0 +1,426 @@
...
+            min: 0,
+            max: max,
+            step: step,
+            mode: 'slider',
+        } as const
+        config['components'][name] = comp
+
+        /*
+         * Setpoint is hours×60 (minutes) in tlv, same as RAC timers on 0x21b.
+         * While counting down, notifies send remaining time in seconds.
```
anszom: See the other comment about timers, this is also confusing. If the HA unit is correct (hours), then read_xform & write_xform would translate to/from minutes on the TLV side.

anszom (commented):
Sorry for the mess, I submitted some comments out-of-review. Only a few minor details to fix. Also please add a README entry with the device name.

tests/cloud/devices/DHUM_056905_WW.test.ts
BluSyn: Leftover from refactor, easy rename.

cloud/homeassistant.ts
BluSyn: Was used before, re-added this

BluSyn:
@anszom Updated comments and few bits to resolve issues you highlighted. Also rebased again due to conflicts on master. Confirmed latest changes are working on local hardware.

anszom:
Did a manual squash-merge. Closing the PR.

Thank you for your patience :)

PR #71, opened 2026-06-13
Trivial: RAC_056905_WW: add state_class HA attribute where appropriate

maciejsszmigiero:
Home Assistant requires this attribute in order to generate long-term statistics for these sensors.

Adding such attribute also makes HA treat EEV opening value as a number and create an appropriate chart for it, instead of treating it as a free-form text that can't be charted.


anszom:
:+1:

PR #82, opened 2026-06-28
feat: add RAC_0B0001_WW as alias for RAC_056905_WW

oirad:
`RAC_0B0001_WW` is a European variant of the DualCool wall-mount unit (`deviceType: 401`, `RTK_RTL8720cm`, `protocolVer: 4.9`, reports `countryCode: DE`, deployed in Italy). It uses the same TLV packet format as `RAC_056905_WW` and the existing handler works without modification.

Confirmed across five units (models EZ09CSNCSJ1, EZ12CSUCA31, EZ12CSNNCSJ1). All climate entities (power, mode, target temp, fan speed, swing) work correctly in Home Assistant.

The device connects, sends `preDeploy`, completes provisioning, then streams `device_packet` telemetry continuously:

```
incoming clip/provisioning/devices/<uuid>  cmd: preDeploy  kind: RAC_0B0001_WW
incoming lime/devices/<uuid>               cmd: completeProvisioning
incoming clip/message/devices/<uuid>       cmd: completeProvisioning_ack
incoming clip/message/devices/<uuid>       cmd: device_packet  data: "000004000000870204A204..."
incoming clip/message/devices/<uuid>       cmd: device_packet  data: "000004000000870204F304..."
```

Without this alias, rethink logs `thinq2 device type RAC_0B0001_WW unknown` and no HA entities are created.

**Change:** one-line alias in `t2deviceTypes` map — no new handler needed.

### DIFF
--- a/cloud/ha_bridge.ts
+++ b/cloud/ha_bridge.ts
@@ -27,6 +27,7 @@ const t1deviceTypes: Record<string, T1Factory> = {
 const t2deviceTypes: Record<string, T2Factory> = {
     POT_056905_WW,
     RAC_056905_WW,
+    ['RAC_0B0001_WW']: RAC_056905_WW, // European variant (deviceType 401, RTK_RTL8720cm), same TLV handler
     WIN_056905_WW,
     ['2REF11EIDA__4']: Dev_2REF11EIDA__4,
     ['2RES1VE61NFA2']: Dev_2RES1VE61NFA2,


maciejsszmigiero:
> RAC_0B0001_WW is a European variant 

I guess you mean _another_ European variant of DualCool family since the existing unit families that use `RAC_056905_WW` are AFAIK all European market too.

oirad:
> I guess you mean another European variant

Indeed, thanks for precising it :) 

BTW Thanks for the work you did on this, I have to say it works like a charm, if it can be helpful I also made this all into an [addon](https://github.com/oirad/hassio-rethink), that I would be happy to just port in here if you prefer not keeping it separate and if you want I can open another PR in case.

anszom:
> BTW Thanks for the work you did on this, I have to say it works like a charm, if it can be helpful I also made this all into an [addon](https://github.com/oirad/hassio-rethink), that I would be happy to just port in here if you prefer not keeping it separate and if you want I can open another PR in case.

I wanted to sort out some issues with provisioning & port configuration before making an "official" addon, but it seems we're almost there. Feel free to open a PR for that.

(I've fixed the comment & merged this PR, closing)

PR #87, opened 2026-06-30
Add ThinQ2 device handler for LG F4X7511TWS washer (VCDWL2QEUK)

maslygan:
Adds support for the LG F4X7511TWS front-load washer (internal model `VCDWL2QEUK`, ThinQ2). Unlike the single-frame F-class washers, this model emits several AA/BB frame types, so it gets its own handler.

## What this adds

A device handler (`cloud/devices/VCDWL2QEUK.ts`) parsing the AA/BB-framed packets to expose:

| Entity | Type | Source |
|---|---|---|
| Power | binary_sensor (`running`) | frame-derived (see below) |
| Status | sensor | `0x92` record B, step code `rec[20]` |
| Soil level | sensor | `0x92` record B `rec[0]` |
| Course | sensor | `0x92` record B `rec[4]` |
| Temperature | sensor (°C) | `0x92` record B `rec[1]` |
| Spin | sensor (RPM) | `0x92` record B `rec[3]` |
| Rinse | sensor | `0x92` record B `rec[26]` |
| Energy | sensor (Wh, `total_increasing`) | `0x92` record B `rec[16:17]` big-endian |
| Initial time | sensor (duration) | `0x92` record B `rec[15]` |
| Remaining time | sensor (duration) | `0x92` record B `rec[13]` |
| Detergent / Softener dispenser | binary_sensor | `0x92` record B `rec[29]` / `rec[30]` (EzDispense) |
| Detergent / Softener dose | sensor (mL) | `0x92` record B `rec[31]` / `rec[32]` |
| Pre-wash / TurboWash / Steam | binary_sensor | `0x92` record B `rec[33]` / `rec[34]` bitflags |
| Child lock / Remote start | binary_sensor (diagnostic) | `0x92` record B `rec[36]` bits `0x20` / `0x10` |
| Washes since drum clean | sensor (diagnostic) | `0x92` record B `rec[27]` (LG `TCLCount`) |
| Door | binary_sensor (`door`) | `0x41` snapshot `buf[18]` |

## How it works

Frame types are discriminated by `inner[3]`:

- `0x88` config — sent at cycle start / setting change; used only as an early power-ON signal.
- `0x92` status — carries two back-to-back 63-byte records; the current state is record B (`buf[78:141]`). cf. #77's dual-record `0xEC` washer, which uses the *first* record — this model uses the second (verified against the live app countdown).
- `0x41` door snapshot — door state at `buf[18]` (`0x02` = closed, `0x01` = open).

Record B uses the **same offsets in every phase** of a cycle. During the wash phase it leads with the soil level (light/medium/heavy); during the rinse/spin/end phases and the spin-only / rinse+spin programs it leads with `0x00` (no wash intensity) but still carries a non-zero status byte. Power is frame-derived (no timers): a record with **both** a zero soil lead and a zero status byte is standby → OFF; any non-zero status is a running frame → ON. Settings (soil, course, temp, spin, rinse, the EzDispense doses, the program-option toggles, the child-lock/remote-start flags and the tub-clean counter) are read from the periodic status frame rather than the sparse `0x88` config frame, which is routinely missed.

## Design notes (intentional differences from the F-class washers)

This handler follows the leaner, read-only shape of #77 (the other dual-record washer) rather than the command-capable F-class handlers:

- **Read-only.** All entities are sensors — this model's control/command frames haven't been reverse-engineered, so there's no writable power `switch`, `start`/`pause`, or `start()` init query (no untested commands are sent to the appliance).
- **`status`, `course`, `soil` and `rinse` are plain sensors, not `device_class: enum`.** The codes are model-specific and the handler emits an `unknown`/`Running` fallback; HA rejects values outside an enum's fixed `options`, so free-text sensors are used.
- **`child_lock` uses an icon, not `device_class: lock`** — HA's `lock` binary_sensor class is inverted (`on` = unlocked), which would flip the meaning. `child_lock` and `remote_start` are physical-panel flags and are marked `entity_category: diagnostic`.
- **`door` is the physical open/closed contact** (`device_class: door`, from the `0x41` snapshot). A separate door-lock state isn't exposed — it doesn't map to a clean bit on this model, and is essentially "running and not paused", which `status` already conveys.
- **`Washes since drum clean`** is the LG `TCLCount` (a tub-clean reminder counter that resets at each drum clean); it's a plain sensor with **no `state_class`**, since `total_increasing` would wrongly accumulate across the reset. This model doesn't expose a lifetime cycle count.
- **Lean entity set.** Only decoded fields are exposed; `error` and the raw control state are omitted rather than published as a stuck `unknown`.

## Known limitations

- `error` isn't located yet — that entity is intentionally omitted rather than published as a stuck `unknown`.
- A few cloud-reported fields aren't mapped to wire bytes yet (door-lock, load/weight detection, EzDispense empty/refill state); each has only a single time-confounded transition per cycle, so they'd need targeted captures.
- The power-OFF signature is validated for standby-off after a normal program; off-states after other program families haven't been captured yet.
- The `0x41` door snapshot is only emitted while powered on, so the door entity retains its last value during standby.

Marked as mostly-working in the README.

## How this was built

This handler was primarily written by Claude (Opus 4.8, via Claude Code) and tested thoroughly against real frames captured from a live machine: power on/off, the full run of status phases (detecting → wash → rinse → spin → end), soil, course/temp/spin, rinse, energy, the EzDispense dispensers and doses, the program-option toggles, the child-lock and remote-start flags, the tub-clean counter, and the door were each validated on hardware as well as in unit tests. Field offsets were cross-checked against the LG cloud's own decode of the same frames. I've also cross-checked the existing device PRs to line this up with the project's structure and conventions as closely as possible. Happy to revise anything that doesn't fit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ The following appliances are currently supported in rethink:
     - 👍 F2J7HG1W, Washing Machine - mostly working,
     - 🫤 F4WV709P1E, Front-Loading Washing Machine - preliminary support
     - 🫤 TW4V9RW9W - preliminary support
+    - 🫤 F4X7511TWS (VCDWL2QEUK), Front-Load Washing Machine - preliminary support
 
 The supported appliances can be used "out of the box" with HomeAssistant or another compatible MQTT consumer.  
 Appliances not listed above can still be used with the bridge mode, but they will not be translated to MQTT. Contributions are welcome!
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ The following appliances are currently supported in rethink:
     - 👍 F2J7HG1W, Washing Machine - mostly working,
     - 🫤 F4WV709P1E, Front-Loading Washing Machine - preliminary support
     - 🫤 TW4V9RW9W - preliminary support
-    - 🫤 F4X7511TWS (VCDWL2QEUK), Front-Load Washing Machine - preliminary support
+    - 👍 F4X7511TWS (VCDWL2QEUK), Front-Load Washing Machine - mostly working
 
 The supported appliances can be used "out of the box" with HomeAssistant or another compatible MQTT consumer.  
 Appliances not listed above can still be used with the bridge mode, but they will not be translated to MQTT. Contributions are welcome!
--- a/cloud/devices/VCDWL2QEUK.ts
+++ b/cloud/devices/VCDWL2QEUK.ts
@@ -22,20 +22,21 @@ const CONFIG_FRAME_TYPE = 0x88
 const STATUS_RECORD_OFF = 78 // record B (current state) start within the inner body
 const STATUS_RECORD_LEN = 63
 const STATUS_FRAME_LEN = 142 // 15B header + 63B recA + 63B recB + 1B trailer
-// rec[0] of record B = the wash-intensity (soil) level while running; 0x00 when standby/spin-only.
-// So rec[0] doubles as the running-vs-standby discriminator (any non-zero soil level = running). RE'd
-// by stepping the soil level on a fixed course and diffing record B (each level confirmed twice).
+// rec[0] of record B = the wash-intensity (soil) level while washing; 0x00 during the rinse/spin/end
+// phases, the spin-only program, and standby (there is no wash intensity then). RE'd by stepping the soil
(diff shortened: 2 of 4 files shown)


anszom:
All looks good, thanks!

PR #88, opened 2026-07-04
Trivial: RAC_056905_WW HA climate.turn_on command fix

maciejsszmigiero:
Running Home Assistant `climate.turn_on` command on `RAC_056905_WW` climate component wasn't actually turning the device on.

The issue seems to be due to missing temperature tag (`0x1fe`) on the power on command - the device mode + fan speed + temperature tag triple should always go together.

Writing 1 to the power tag (`0x1f7`) alone does not seem to do anything and in fact seems to be ignored by the device.

Add the missing `0x1fe` tag to `write_attach` when turning climate power ON to fix this issue.

Also, opportunistically remove `write_attach [0x1f9, 0x1fa]` from swing modes - there's no real reason why changing swing modes would require attaching device mode and fan speed tags and the IR remote does not seem to send device mode and fan speed with swing setting either.

### DIFF
--- a/cloud/devices/RAC_056905_WW.ts
+++ b/cloud/devices/RAC_056905_WW.ts
@@ -401,7 +401,6 @@ export default class Device extends TLVDevice {
                     }
                     return modes2clip[val]
                 },
-                write_attach: [0x1f9, 0x1fa],
             })
         }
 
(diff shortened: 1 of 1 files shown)


anszom:
I was sure that it had worked correctly, but - sure enough - my AC behaves exactly as you've described. Thanks for the fix.

PR #91, opened 2026-07-12
Add ThinQ2 device handlers for WM3900HBA washer and DLEX3900B dryer

BatemanJo9:
## Summary
- Adds `cloud/devices/F3L2CYU__.ts` for the LG WM3900HBA washer (self-reported ThinQ2 modelId `F3L2CYU__`) — decodes course, phase, remaining/reserve time, soil, spin, temp, all six option toggles (extra rinse, pre wash, steam, cold wash, turbo wash, delay wash), and door-closed/door-lock state.
- Adds `cloud/devices/RV13B6BSD_D_US_WIFI.ts` for the companion LG DLEX3900B dryer — decodes course, dry level, temp, time, all six cycle phases, Reduce Static, Energy Saver, Turbo Steam, Wrinkle Care, Child Lock, and Damp Dry Signal.
- Fixes the washer handler to also decode the single-record `0xEB` frame sent right after reconnect (before the dual-record `0xEC` frames arrive), so Home Assistant entities get real values immediately instead of sitting empty until the first `0xEC`.

Every decoded field/offset was live-verified against the LG cloud's own decoded state via the `rethink-agent` MCP tooling while operating the physical appliances (dial browsing, individual setting toggles, full wash/dry cycles, pause/resume, and remote start/pause/power-off from the official LG app) — not inferred from static analysis alone. Both handlers have been running in daily production use against the real appliances through Home Assistant for a multi-day soak period with no issues.

## Test plan
- [x] Unit tests added for both handlers using real captured frames for each decoded field (`tests/cloud/devices/F3L2CYU__.test.ts`, `tests/cloud/devices/RV13B6BSD_D_US_WIFI.test.ts`)
- [x] Deployed live against real appliances, confirmed working end-to-end through Home Assistant across multiple real wash/dry cycles

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,8 @@ The following appliances are currently supported in rethink:
     - 🫤 TW4V9RW9W - preliminary support
     - 👍 F4X7511TWS (VCDWL2QEUK), Front-Load Washing Machine - mostly working
     - 👍 WM3900HBA (F3L2CYU\_\_), Front-Load Washing Machine - mostly working
+- Dryers:
+    - 👍 DLEX3900B (RV13B6BSD_D_US_WIFI), Electric Dryer - mostly working
 
 The supported appliances can be used "out of the box" with HomeAssistant or another compatible MQTT consumer.  
 Appliances not listed above can still be used with the bridge mode, but they will not be translated to MQTT. Contributions are welcome!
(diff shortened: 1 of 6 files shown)


anszom:
Looks good, thanks!

BatemanJo9:
And thank you!!!

PR #96, opened 2026-07-17
Initial Support for GML844PZKZ Multi-Door Slim Refrigerator

flecke-m:
Adding device 2REF12EII_P_2 to ha_bridge.ts and its own file.

Supported settings are freezer temperature, fridge temperature, express freeze and Pure N Fresh filter settings. Additionally door open/close is working as well.

Every functionality has been tested through bridge mode comparing the LG App vs. Home Assistant.

Added test file to complete the pull request.


anszom:
Does this model support Fahrenheit temperatures? if so, then it would be good to include support for this as well.

flecke-m:
I can see a Fahrenheit Symbol in the Display, but I have double checked the ThinQ App and as well conducted the manual, both do not show a Fahrenheit support. The fridge itself also has on the full model number the suffix EUR, which most likely stands for Europe. So I assume it is not supported.

So far I investigated all values with your tool to comment and intercept the traffic within bridge mode. I will also do a second round of investigation with ingesting data to the cloud and see what happens, but currently I ran a bit out of time.

anszom:
I've squashed your commits to branch 2REF12EII_P_2. I've also removed the references to some of your development files, I hope you don't mind.

I'd like to make some changes before merging. I was hoping you could make these because I don't have a device to verify against.

1. The fridge status block appears to match `packStatus` & `unpackStatus` from `fridge_common.ts`. This would clarify the "ack flag" which is actually the temperature unit. It would be better to use these common functions instead of managing byte offsets manually.
2. The door state should be extracted from the status block as well, this way it will be correct at startup.
3. The _F017 command buffers in the test file are not referenced (this was made more obvious when I removed the `captures()` wrapper). On top of that, one of them is one byte longer than the header would suggest, and doesn't seem to match the freezer setpoint write at all. Please re-check this - and - preferably - rewrite the test to match against the captured packet.

flecke-m:
Thank you, I will have a more thorough look at it with the background you have shared. I will then commit the code to the separate branch you created. It is very helpful.

PR #102, opened 2026-07-25
Add WashTower (`WTL_FXU_BDV_NA_01`)  support

schmittjoseph:
Add support for the `WTL_FXU_BDV_NA_01` WashTower (a combined washer+dryer unit exposed as a single device).

Notably:
- Properties are namespaced by unit (`washer/…`, `dryer/…`), requiring a new `+/+/+/set` MQTT subscription for command routing alongside the existing `+/+/set`.
- Door state is delta-only (the device only emits door events on change, never in full state resyncs). `start()` defaults both doors to `CLOSE` so HA has a sensible initial state.

This isn't complete, but I've documented my current understanding of the protocol, even for sensors not yet exposed. Notably the largest missing functionality here is:
- The ability to configure what courses are available on-device as they are "downloaded" (read enabled) via the LG cloud
- The ability to remotely send and start a course. I've done some of the initial work of adding support for configuring remote start pinning (`remote_maintain`), but haven't gotten around to sending courses yet.
- Exposing the detergent and fabric softener  EZDispense sensors, as the unit allows configuring configuring both to hold detergent or fabric softener. I've documented the relevant sensor fields but haven't exposed them yet until I have a better understanding of how they interact in this combined mode.


cloud/devices/WTL_FXU_BDV_NA_01.ts
```diff
@@ -0,0 +1,1059 @@
...
+                        ],
+                    },
+                },
+            }),
+        )
+    }
+
+    start() {
+        //
+        // Door state is delta-only (0x42/0x4e events), so default to CLOSE on each connect.
```
schmittjoseph, line 863: @anszom the largest problem with this comes from the fact that once the common `deployInterval` is reached, `ha_bridge` will tear down a  device and then fully re-initialize it once it reconnects. This isn't really a problem for devices whose sensors are fully reported during the initial state resync, but for devices with delta-only sensors like this WashTower it means a loss of data. I've opted here to assume the doors are closed by default instead of an indeterminate state to help with HA automations.

I took a look ha_bridge, and it seems reasonable that perhaps (outside of this PR) we  have some kind of reconnection concept where we don't fully tear down a device once the `deployInterval` is reached, but instead put it in an intermediate state. In this state if the device reconnects in a reasonable amount of time, we move it back to the active and so MQTT subscribers (HA) are none the wiser to this implementation detail and no delta-only data is lost. If the device doesn't reconnect in a reasonable time then we fully tear it down. I'm curious your thoughts on this, or if you see potential issues here.

It's definitely possible that there's some way to deterministically get the door states, but I've yet to find it so this isn't super pressing unless more devices start having delta-only sensors.

anszom: I'm not sure if I follow. To be honest, I've had to check the sources to find the `deployInterval` value, I have no idea what the device does with it - mine doesn't seem to act on it. Does your device do some kind of periodic reconnect?

schmittjoseph: From what I can tell, at least for this device, the device seems to abruptly close its connection to rethink every ~10 minutes if idle (it seems to correspond to the `deployInterval`, but I havent thoroughly tested yet) and then re-establishes a new connection which results in the door sensor state being lost since rethink fully tears down and then re-initializes the device. I haven't dug too deeply into it yet since I was able to work around it for my needs, so no worries if you're not sure.

anszom: This doesn't seem normal. I guess that the device expects the backend to send something and times out -> reconnects.

schmittjoseph: I had a chance to go through my notes and logs again (it's been a month since I've closely looked at this). I'm going to close this topic since it appears the device reconnection only happens after several hours now instead of every 10 minutes which seems much more reasonable (Every ~10 minutes was happening when I was first working on the device, but I haven't been able to repro it since). If I see it again I'll look more into it, but until then I'll mark this as resolved.

cloud/devices/WTL_FXU_BDV_NA_01.ts
```diff
@@ -0,0 +1,1059 @@
...
+        // Known header structure (common to all packet types):
+        //   [0..2] = 36 0a 00  (fixed prefix)
+        //   [3]    = message type discriminator
+        //   [4]    = 00        (unknown)
+        //   [5]    = unknown
+        //   [6]    = sequence counter
+        //   [7..12] = unknown
+
+        // Strip the common 13-byte header and trailing checksum
+        const body = buf.subarray(FIXED_HEADER_LENGTH, buf.length - 1)
```
anszom, line 924: ProcessAABB already strips the typical xor checksum. Does this packet contain another checksum on top of that?

cloud/devices/WTL_FXU_BDV_NA_01.ts
```diff
@@ -0,0 +1,1059 @@
...
+        this.publishProperty('washer/load_level', Device.formatEnum(WASHER_LOAD_LEVEL, block[26]))
+        this.publishProperty('washer/rinse_count', Device.formatEnum(WASHER_RINSE_COUNT, block[29]))
+        this.publishProperty('washer/laundry_texture', block[43]) // Reported as integer value
+
+        //
+        // Both washer and dryer have an init_lcd, and the command appears to support per-device config
+        // but in real world testing it doesnt appear possible for them to have different values.
+        // Use the washer's as the source of truth.
+        //
+        this.publishProperty('shared/init_lcd', Device.formatEnum(INIT_LCD_THEMES, block[48]))
```
anszom, line 1031: `buzzer` and `init_lcd` have a fixed `options` list. Publishing an out-of-list value via `formatEnum` will trigger HA warnings. For these fields, you should either skip the publish altogether, or publish a value of `unknown` which is universally understood by HA.

anszom (changes requested):
Just a few minor changes before we get this merged.

cloud/devices/WTL_FXU_BDV_NA_01.ts
schmittjoseph: Updated this comment to better reflect my understanding. It seems to act like an extra checksum but I haven't been able to figure out the algorithm for it. It could also be a non-obvious sensor field I haven't mapped yet but I haven't found any proof of that yet.

If I figure it out later I'll open a PR to clarify this.

anszom:
Thanks. I've reshaped the commits a bit (hope you don't mind) & merged the code into master.

schmittjoseph:
> Thanks. I've reshaped the commits a bit (hope you don't mind) & merged the code into master.

That's fine with me. I'd also be fine with a squash merge in the future if that makes it easier on you since it requires no manual rewriting of commits.

PR #103, opened 2026-07-25
RAC_056905_WW: Add fan RPM HA sensor

maciejsszmigiero:
Tag `0x331` is the IDU current fan RPM, this PR adds its HA sensor.


anszom:
:+1:

Issue #105, opened 2026-07-28
Plan: common AC logic into ac_common


3735943886:
Will add an ac_common module. Current RAC logic will be updated so that common functions are shared, leaving only model-specific settings to be configured individually.

For example, `RAC_056905_WW` will be updated as follows:
```typescript
import ACDevice, { type WireLevels } from './ac_common'

/*
 * LG DualCool residential air conditioners (deviceType 401), the model this handler was
 * developed against being RAC_056905_WW. The protocol handling lives in ac_common; below are
 * the properties of these units: the 0x1f9 / 0x1fa scales, the setpoint range, how their power
 * reading has to be corrected, and the fact that they carry a basic-filter counter.
 */
export default class Device extends ACDevice {
    /* TODO: get allowed op modes from 0x2c1 */
    readonly modeLevels: WireLevels = [
        ['cool', 0],
        ['dry', 1],
        ['fan_only', 2],
        ['heat', 4],
        ['auto', 6],
    ]

    /* TODO: get from 0x2c2 */
    readonly fanLevels: WireLevels = [
        ['auto', 8],
        ['very low', 2],
        ['low', 3],
        ['medium', 4],
        ['high', 5],
        ['very high', 6],
    ]

    /* Air purify and energy saving do not read back reliably here */
    readonly modeDependentSwitchOptimistic = true

    /* TODO: some devices report the temp ranges via tags 0x2e1 - 0x2ec */
    temperatureRange() {
        return { min: 18, max: 30 }
    }

    /* Basic filter management, unless the unit reports the unsupported mFilter one */
    filterStyle() {
        return !(this.raw_clip_state[0x2f1] & 1 || this.raw_clip_state[0x2f1] & 0x200)
            ? ('priv' as const)
            : ('none' as const)
    }

    /* This value is reported as zero by multi-split units */
    hasPowerSensor() {
        return !!this.raw_clip_state[0x2b3]
    }

    /*
     * The measurements reported by AC appear to be Watts, but they are not accurate in several aspects:
     * - the value is biased by +50
     * - idle consumption (around 4W) and the 4-way valve is not included
     * - fan modes' consumption appears to be approximated
     *
     * The formula below is expected to be within +/-10% of the actual power consumption. The discrepancy may
     * be highest in fan-only modes.
     */
    powerReadXform(raw: number) {
        return Math.max(5, raw - 60)
    }
}
```

Any comments are appreciated.

kkqq9320:
Thanks for doing this.

[...]

PR #111, opened 2026-08-02
Add nine Korean LG appliances, an ac_common base, and persistent driver stores

tunarabuuu:
This adds support for nine Korean-market LG appliances I've been running daily through rethink, each verified against a live unit with bridge mode on (so every mapping had three witnesses: the value HA set, the tag the appliance reported, and `lge_thinq`'s decode of the same appliance), and each covered by decode-vector regression tests.

## Credit where it's due

**`DHUM_056905_WW` builds directly on @BluSyn's #64** — that driver, the `tlv_device` 0xA7 acceptance, and the `HumidifierComponent` discovery type all originate there (the in-code comments say so too). This PR's version extends it: properties beyond the panel's reach mapped from the model JSON, English labels, and a `ModeTables` export that the newer `DHUM_231006_WW` generation reuses. **If #64 lands first I'll gladly rebase on top of it** — it deserves to land on its own merits.

## New device handlers

| model | appliance | deviceType | notes |
|---|---|---|---|
| `CST_170004_WW` | wall/built-in DualCool IDU | 401 | on a new **`ac_common`** base |
| `DHUM_056905_WW` | dehumidifier | 402 | builds on #64, see above |
| `DHUM_231006_WW` | dehumidifier | 402 | RTL8720cm generation, reuses the 056905 tables |
| `HUM_056905_WW` | PuriCare humidifier | 404 | wire map from LG's own modelJSON `tlv_*` labels |
| `ST_B_E4H01Y_APL` | Styler S5BBP | 203 | AA FF frame layout |
| `HWWA9K_F2` | CordZero A9 stick vacuum | 504 | 12-field AA..BB frame |
| `1WPU4CIGCR__2` | water purifier ATOM-U | 103 | with local usage history |
| `FAKPK21021` | WashTower washer (KR) | 221 | dedicated 0xEB/0xEC decoder |
| `BDH_D39301_KR` | WashTower dryer (KR) | 222 | shared `washtower_common` base |

**About `ac_common`**: in #101 you suggested splitting a shared AC base class instead of translating packets to RAC's expectations — this PR contains that split, and `CST_170004_WW` is its first user. @3735943886's `CST_570004_WW` ceiling cassette should be able to sit on the same base.

The Korean WashTower models are complementary to #102's `WTL_FXU_BDV_NA_01` (different generation, different frame format — these speak a dedicated 0xEB/0xEC encoding decoded in `washtower_common`).

## Infrastructure the drivers stand on

- **`TLVDevice`**: byte 6 may be `0xA7` as well as `0x87` — the dehumidifier family frames with `0xA7` (from #64).
- **Persistent driver stores**: a `ReservationStore` (absolute reservation deadlines survive restarts, used by the AC) and a `PurifierHistoryStore` (water-purifier usage history), both small JSON stores threaded through the bridge as optional constructor parameters. Without a bridge storage path the drivers run stateless.
- **Discovery types**: `HumidifierComponent` (from #64), plus `modes`/`preset_modes` on `ClimateComponent`.
## Verification

- `npm test` — **502 passed, 0 failed** (251 upstream + 251 new)
- `tsc -p tsconfig.build.json --noEmit` — clean
- `prettier --check` — clean

Independent of #110 (only trivially-mergeable overlap in `cloud/homeassistant.ts` type declarations). Happy to split this into per-device PRs if that's easier to review.

### DIFF
--- a/bridge/purifier-history-store.ts
+++ b/bridge/purifier-history-store.ts
@@ -1,0 +1,323 @@
+import { readFileSync, readdirSync, renameSync, writeFileSync } from 'node:fs'
+import { join } from 'node:path'
+import { validDeviceId } from '@/util/device-id'
+
+export type PurifierUsage = {
+    day: string
+    normal: number
+    hot: number
+    cold: number
+    soda: number
+    mineral: number
+    sterilization: number
+}
+
+export type PurifierDailyUsage = Omit<PurifierUsage, 'day'>
+
+export type PurifierSterilizationHistory = {
+    active?: {
+        kind: 'pipe' | 'outlet'
+        startedAt: number
+    }
+    lastPipeAt?: number
+    lastOutletAt?: number
+    countsByMonth: Record<string, number>
+}
+
+export type PurifierScheduleAnchor = {
+    parts: [number, number, number, number]
+    instant: number
+}
+
+export type PurifierHistoryState = {
+    version: 1
+    collectedSince: number
+    usageComplete: boolean
+    usage: PurifierUsage
+    sterilization: PurifierSterilizationHistory
+    schedule?: PurifierScheduleAnchor
+}
+
+export type PurifierHistoryStore = {
+    load(id: string): PurifierHistoryState | undefined
+    save(id: string, state: PurifierHistoryState): void
+}
+
+type PurifierHistoryFileOps = {
+    read(path: string): string
+    write(path: string, data: string, options: { mode: number }): void
+    rename(from: string, to: string): void
+}
+
+type PurifierHistoryWarning = (message: string, error: unknown) => void
+
+const fileOps: PurifierHistoryFileOps = {
+    read: (path) => readFileSync(path, 'utf-8'),
+    write: (path, data, options) => writeFileSync(path, data, options),
+    rename: (from, to) => renameSync(from, to),
+}
+
+const usageFields = ['normal', 'hot', 'cold', 'soda', 'mineral', 'sterilization'] as const
+
+function isRecord(value: unknown): value is Record<string, unknown> {
+    return value != null && typeof value === 'object' && !Array.isArray(value)
+}
+
+function isNonnegativeSafeInteger(value: unknown): value is number {
+    return typeof value === 'number' && Number.isSafeInteger(value) && value >= 0
+}
+
+function isTimestamp(value: unknown): value is number {
+    return typeof value === 'number' && Number.isFinite(value) && value >= 0
+}
+
+function validDay(day: unknown): day is string {
+    if (typeof day !== 'string') return false
+    const match = /^(\d{4})-(\d{2})-(\d{2})$/.exec(day)
+    if (!match) return false
+
+    const year = Number(match[1])
+    const month = Number(match[2])
+    const dayOfMonth = Number(match[3])
+    if (month < 1 || month > 12) return false
+
+    const leapYear = year % 4 === 0 && (year % 100 !== 0 || year % 400 === 0)
+    const daysInMonth = [31, leapYear ? 29 : 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31]
+    return dayOfMonth >= 1 && dayOfMonth <= daysInMonth[month - 1]!
+}
+
+function validMonth(month: string) {
+    const match = /^(\d{4})-(\d{2})$/.exec(month)
+    if (!match) return false
+    const monthNumber = Number(match[2])
+    return monthNumber >= 1 && monthNumber <= 12
+}
+
+function canonicalSchedule(value: unknown): PurifierScheduleAnchor | undefined {
+    if (!isRecord(value) || !Array.isArray(value.parts) || value.parts.length !== 4 || !isTimestamp(value.instant))
+        return undefined
+
+    const parts = value.parts
+    if (parts.some((part) => typeof part !== 'number' || !Number.isSafeInteger(part) || part < 0 || part > 0xff))
+        return undefined
+
+    const [month, day, hour, minute] = parts as [number, number, number, number]
+    const date = new Date(value.instant)
+    if (
+        date.getUTCMonth() + 1 !== month ||
+        date.getUTCDate() !== day ||
+        date.getUTCHours() !== hour ||
+        date.getUTCMinutes() !== minute ||
+        date.getUTCSeconds() !== 0 ||
+        date.getUTCMilliseconds() !== 0
+    )
+        return undefined
+
+    return { parts: [month, day, hour, minute], instant: value.instant }
+}
+
+function canonicalState(value: unknown): PurifierHistoryState | undefined {
+    if (!isRecord(value) || value.version !== 1 || !isTimestamp(value.collectedSince)) return undefined
+    if (!isRecord(value.usage) || !isRecord(value.sterilization)) return undefined
+
+    if (!validDay(value.usage.day)) return undefined
+    const usage = {
+        day: value.usage.day,
+    } as PurifierUsage
+    if (value.usageComplete !== undefined && typeof value.usageComplete !== 'boolean') return undefined
+    for (const field of usageFields) {
+        const amount = value.usage[field]
+        if (!isNonnegativeSafeInteger(amount)) return undefined
+        usage[field] = amount
+    }
+
+    const storedSterilization = value.sterilization
+    if (!isRecord(storedSterilization.countsByMonth)) return undefined
+
+    const countsByMonth: Record<string, number> = {}
+    for (const [month, count] of Object.entries(storedSterilization.countsByMonth)) {
+        if (!validMonth(month) || !isNonnegativeSafeInteger(count)) return undefined
+        countsByMonth[month] = count
+    }
+
+    const sterilization = {} as PurifierSterilizationHistory
+    if (storedSterilization.active !== undefined) {
+        if (
+            !isRecord(storedSterilization.active) ||
+            (storedSterilization.active.kind !== 'pipe' && storedSterilization.active.kind !== 'outlet') ||
+            !isTimestamp(storedSterilization.active.startedAt)
+        )
+            return undefined
+        sterilization.active = {
+            kind: storedSterilization.active.kind,
+            startedAt: storedSterilization.active.startedAt,
+        }
+    }
+    if (storedSterilization.lastPipeAt !== undefined) {
+        if (!isTimestamp(storedSterilization.lastPipeAt)) return undefined
+        sterilization.lastPipeAt = storedSterilization.lastPipeAt
+    }
+    if (storedSterilization.lastOutletAt !== undefined) {
+        if (!isTimestamp(storedSterilization.lastOutletAt)) return undefined
+        sterilization.lastOutletAt = storedSterilization.lastOutletAt
+    }
+    sterilization.countsByMonth = countsByMonth
+
+    const state: PurifierHistoryState = {
+        version: 1,
+        collectedSince: value.collectedSince,
+        // Version 1 files written before cloud baselining cannot prove that the bridge
+        // observed the whole KST day. Treat them as partial instead of publishing a false zero.
+        usageComplete: value.usageComplete === true,
+        usage,
+        sterilization,
+    }
+    if (value.schedule !== undefined) {
+        const schedule = canonicalSchedule(value.schedule)
+        if (!schedule) return undefined
+        state.schedule = schedule
+    }
+    return state
+}
+
+export class PurifierHistoryJSONStore implements PurifierHistoryStore {
+    private tempSequence = 0
+
+    constructor(
+        readonly basePath: string,
+        private readonly files: PurifierHistoryFileOps = fileOps,
+        private readonly warn: PurifierHistoryWarning = console.warn,
+    ) {}
+
+    private path(id: string) {
+        if (!validDeviceId(id)) return undefined
+        return join(this.basePath, `purifier_${id}.json`)
+    }
+
+    private stateForDay(id: string, day: string, now: number, reportLoadFailure = true) {
+        const existing = reportLoadFailure ? this.load(id) : this.loadWithoutWarning(id)
+        if (existing) {
+            return {
+                ...existing,
+                usage:
+                    existing.usage.day === day
+                        ? { ...existing.usage }
+                        : {
+                              day,
+                              normal: 0,
+                              hot: 0,
+                              cold: 0,
+                              soda: 0,
+                              mineral: 0,
+                              sterilization: 0,
+                          },
+            }
+        }
+        return {
+            version: 1 as const,
+            collectedSince: now,
+            usageComplete: false,
+            usage: {
+                day,
+                normal: 0,
+                hot: 0,
+                cold: 0,
+                soda: 0,
+                mineral: 0,
+                sterilization: 0,
+            },
+            sterilization: { countsByMonth: {} },
+        }
+    }
+
+    markAllDailyUsageIncomplete(day: string, now: number) {
+        let filenames: string[]
+        try {
+            filenames = readdirSync(this.basePath)
+        } catch (error) {
+            this.warn('Unable to enumerate purifier history for daily usage preparation', this.safeError(error))
+            return []
+        }
+        const deviceIds: string[] = []
+        for (const filename of filenames) {
+            const match = /^purifier_(.+)\.json$/.exec(filename)
+            if (!match || !validDeviceId(match[1]!)) continue
+            deviceIds.push(match[1]!)
+            try {
+                const state = this.stateForDay(match[1]!, day, now, false)
+                state.usageComplete = false
+                this.save(match[1]!, state)
+            } catch (error) {
+                this.warn('Unable to prepare one purifier history for daily usage', this.safeError(error))
+            }
+        }
+        return deviceIds
+    }
+
+    private safeError(error: unknown) {
+        return { error: error instanceof Error ? error.name : 'UnknownError' }
+    }
+
+    private loadWithoutWarning(id: string) {
+        const path = this.path(id)
+        if (!path) return undefined
+        let serialized: string
+        try {
+            serialized = this.files.read(path)
+        } catch (error) {
+            if ((error as NodeJS.ErrnoException)?.code === 'ENOENT') return undefined
+            throw error
+        }
+        const state = canonicalState(JSON.parse(serialized) as unknown)
+        if (!state) throw new Error('purifier history has an invalid schema')
+        return state
+    }
+
+    prepareDailyUsage(id: string, day: string, now: number) {
+        const state = this.stateForDay(id, day, now, false)
+        state.usageComplete = false
+        this.save(id, state)
+    }
+
+    applyDailyUsageBaseline(id: string, day: string, usage: PurifierDailyUsage, now: number) {
+        const state = this.stateForDay(id, day, now, false)
+        state.usageComplete = true
+        state.usage = { day, ...usage }
+        this.save(id, state)
+    }
+
+    load(id: string): PurifierHistoryState | undefined {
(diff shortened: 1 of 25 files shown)


maciejsszmigiero:
> publishes platform-only tombstone components so an installation that previously ran a raw-capture build has the leftover diagnostic entities cleaned up

Was this `raw-capture build` upstream code or someone's development code?

If that code was never upstream, then we really needn't/shouldn't include cleanup for it upstream.

maciejsszmigiero:
As in #110,  please split this giant commit (**14k** lines changed) into smaller one, one commit per logical change as this is basically non-reviewable in this form.

tunarabuuu:
You're right — it was never upstream code. Those diagnostic entities only ever existed on my own add-on's development builds (and the drivers here graduated out of that setup), so the cleanup belongs on my side, not here.

Removed in the latest push: `dropGenericComponents` and `cloud/devices/generic.ts` are gone, the five drivers publish their discovery config directly, and the tests that asserted the tombstones went with them. Tests are now 502 passed (251 upstream + 251 new), tsc and prettier still clean.

tunarabuuu:
Done — split into 13 commits: four infrastructure pieces first (discovery types and the 0xA7 acceptance, both from #64; the two JSON stores; the bridge threading), then one commit per appliance, each carrying its driver, its tests, and its own registry entry. Every intermediate commit type-checks on its own, and the final tree is byte-identical to the previous push — 502 tests passing.

cloud/devices/korean.ts
```diff
@@ -0,0 +1,19 @@
+import { type DeviceDiscovery } from '../homeassistant'
+
+/**
+ * Rename a device's components into Korean.
+ *
+ * The appliances here are Korean-market units: their own panel, the LG app and the cloud
+ * integration all use Korean, so a driver that labels half its entities in English leaves the
```
maciejsszmigiero: Can't this be done in HA instead? I know one can re-name devices and their individual components in HA.

tunarabuuu: It can, but only for the display names, and one entity at a time. These appliances publish a lot of entities (the WashTower pair alone exposes ~230), so "rename it in HA" means a per-installation manual pass that resets on a fresh install or a re-discovery.

The bigger half can't be renamed in HA at all: the *state values* and select *options* these drivers publish are Korean labels decoded from the appliance's own enums (the same strings the unit's panel and the LG app show — that's what the tests pin). HA renaming covers an entity's name, not the values flowing through it, so without `applyKoreanNames` the device reads half-Korean/half-English no matter what you do in the UI.

So the helper only makes the last bit — component display names — consistent with what the rest of the device already speaks. That said, if you'd rather have English defaults with Korean as an option (or a config switch), I'm happy to restructure it that way.

maciejsszmigiero:
> About ac_common: in https://github.com/anszom/rethink/pull/101 you suggested splitting a shared AC base class instead of translating packets to RAC's expectations — this PR contains that split

I don't see `RAC_056905_WW` being covered to `ac_common` here and that was kind-of the point of having common AC base class.

I also think that introducing `ac_common` is such an invasive and risky change that it should have its own PR implementing issue #105 - probably coordinating with @3735943886 so it works for their devices too.

tunarabuuu:
Good call — `ac_common` is out of this PR entirely.

You're right on both counts: I hadn't converted `RAC_056905_WW` onto the base (so it wasn't actually *shared* yet), and #105 already exists to do this properly, with @3735943886 and @kkqq9320 working out the design against real captures from several deviceType-401 units. Dropping a parallel ac_common in here would have cut across that.

So I removed `ac_common.ts`, `CST_170004_WW`, and the reservation store that only the AC used (its persistence isn't needed by anything else here). This PR is now just the nine non-AC appliances — dehumidifiers, humidifier, styler, vacuum, water purifier, and the WashTower pair — plus the two small shared bits they need (`HumidifierComponent` and the 0xA7 tag acceptance, both from #64). CST will come back as its own PR on top of whatever #105 lands, and I'm glad to bring my CST_170004_WW captures to that thread.

Rebased to 11 commits (one per appliance + the shared pieces), each type-checks on its own; 485 tests passing.

cloud/devices/DHUM_231006_WW.ts
```diff
@@ -0,0 +1,73 @@
...
+ *
+ *   0x1f7 power             airState.operation 1        -> ON
+ *   0x253 target humidity   tracked 45 / 55 / 65 exactly on each write
+ *   0x1fa fan speed         airState.windStrength
+ *   0x1f9 mode              airState.opMode
+ *   ionizer / UVnano / bucket light matched airState.* one for one
+ *
+ * Only the two enums differ, so that is all this overrides.
+ *
+ * Labels are Korean because this model is sold in Korea and its panel, the LG
```
maciejsszmigiero, line 18: The `_WW` suffix would suggest this is a **W**orld **W**ide device.

It may be sold with panel localized to particular market, but I wouldn't take it as a guarantee that device identifying as `DHUM_231006_WW` isn't sold in any other market with a different panel language.

cloud/devices/DHUM_231006_WW.ts
```diff
@@ -0,0 +1,73 @@
...
+ * app, and the cloud integration all use these names — matching them keeps one
```
maciejsszmigiero, line 19: The official app and HA integration both separate translations from code.

cloud/devices/DHUM_231006_WW.ts
maciejsszmigiero: Same remark for other `_WW` devices in this PR.

tunarabuuu:
Done — everything is English now, no Korean strings anywhere in these drivers.

You're right that `_WW` isn't a Korea-only guarantee and that translations belong out of the code. The wire enums are still decoded from LG's ko-KR modelJSON/langpack (that's where the codes and their meanings come from), but every published label — device names, state values, and select options — is now an English translation of that, not the Korean text. The `korean.ts` helper is gone, and the tests that used to assert labels were Korean now assert they're English.

If a proper i18n layer (canonical keys + a translation table, like the app/HA integration do) is something you'd want, I'm happy to build that as a follow-up — but this at least gets the defaults right for any market.

Re-pushed: still 11 commits (one per appliance + the shared pieces), each type-checks on its own, 485 tests passing.

cloud/devices/1WPU4CIGCR__2.ts
```diff
@@ -0,0 +1,1031 @@
...
+    PurifierDailyUsage,
+    PurifierHistoryState,
+    PurifierHistoryStore,
+    PurifierUsage,
+} from '@/bridge/purifier-history-store'
+
+/*
+ * LG PuriCare Water Purifier (ATOM-U STS T20 Cold/Hot/Purified/Steril), ThinQ model 1WPU4CIGCR__2, deviceType 103.
+ *
+ * THE COMMAND FRAME, AND WHY IT LOOKED LIKE THERE WASN'T ONE
```
anszom, line 17: That's quite a lot of prose, with references to out-of-repo files, and various terms which aren't immediately obvious (`Config.supportControl`, legacy control path, capability API - I don't know what this refers to, and it isn't necessary to understand the device-side protocol).

Please shorten/remove. Focus on what is needed to understand the protocol.

cloud/devices/1WPU4CIGCR__2.ts
```diff
@@ -0,0 +1,1031 @@
...
+    const period = hour < 12 ? 'AM' : 'PM'
+    return `${hour % 12 || 12}:${pad2(minute)} ${period}`
+}
+const formatKst = (time: number | undefined) => {
+    // This bridge can only attest to completions it observed. An absent timestamp does not
+    // prove the appliance has never completed a cycle before collection began.
+    if (time == null) return 'Not yet observed'
+    const p = kstParts(time)
+    return `${p.year}-${pad2(p.month)}-${pad2(p.day)} ${pad2(p.hour)}:${pad2(p.minute)}`
+}
```
anszom, line 201: A few comments here.

1. Time formatting utilities should be moved to somewhere in util/...
2. Is anything here strictly bound to the KST timezone?
3. If not, then it would be both simpler & more general to simply operate in the user's local time.

cloud/devices/1WPU4CIGCR__2.ts
```diff
@@ -0,0 +1,1031 @@
...
+        this.sterilizationScheduleAnchor = undefined
+        this.sterilizationScheduleObserved = false
+        super.drop()
+    }
+
+    /*
+     * The base class strips the AA/length prefix before handing the body on, but every offset
+     * above is stated against the WHOLE frame — that is how the probes recorded them, and
+     * keeping the two in the same coordinate system is what makes the map checkable. So the
+     * frame is taken here, unstripped.
```
anszom, line 632: You can simply subtract two and now you have offsets relative to the internal AABB payload. No reason to circumvent the regular code path used by all other devices.

cloud/devices/1WPU4CIGCR__2.ts
```diff
@@ -0,0 +1,1031 @@
...
+        // in localUsage above; restore it if the atomic save fails.
+        this.pendingUsage = undefined
+        if (!this.saveHistory(next)) {
+            this.pendingUsage = pendingUsage
+            return false
+        }
+        this.history = next
+        this.publishHistory()
+        return true
+    }
```
anszom, line 587: this, along with some other parts appear to be used only in tests. What is the purpose?

cloud/devices/1WPU4CIGCR__2.ts
```diff
@@ -0,0 +1,1031 @@
...
+
+/** wpState.tempUnit / amountUnit — settings the panel owns; diagnostic only. */
+const TEMP_UNIT: Record<number, string> = { 0: '°F', 1: '°C' }
+const AMOUNT_UNIT: Record<number, string> = { 0: 'oz', 1: 'mL' }
+
+const enumOf = (table: Record<number, string>, raw: number) => table[raw] ?? `Code ${raw}`
+
+export default class Device extends AABBDevice {
+    private readonly legacyPlatformCleanup: DeviceDiscovery
+    private historyStore: PurifierHistoryStore | undefined
```
anszom, line 369: I was trying to avoid the need to store any persistent history in `rethink` - usually these things can be supported by HA, and rethink serves only as a simple protocol translator.

So far I think the purifier can be supported in a stateless manner too. HA can handle deltas/accumulated values (even with reset-to-zero, see for example energy consumption for the RAC_...). Any historical data should similarly be retained by HA instead of rethink.

cloud/devices/HWWA9K_F2.ts
```diff
@@ -0,0 +1,340 @@
...
+import { DeviceDiscovery, type Connection } from '../homeassistant'
+import { type Metadata } from '../thinq'
+import { allowExtendedType } from '@/util/casting'
+import HADevice from './base'
+import AABBDevice from './aabb_device'
+
+/*
+ * LG CordZero A9 Stick Vacuum, ThinQ model HWWA9K_F2, deviceType 504.
+ *
+ * WHERE THE BYTE OFFSETS COME FROM
```
anszom, line 11: Please shorten this.

cloud/devices/BDH_D39301_KR.ts
```diff
@@ -0,0 +1,11 @@
+import { Device as Thinq2Device } from '../thinq2/device'
+import { type Connection } from '../homeassistant'
+import { type Metadata } from '../thinq'
+import WashTowerDevice, { DRYER_MODEL } from './washtower_common'
```
anszom, line 4: From what I can see here, both washtowers are supported in an identical way. If you don't expect any differences to appear, it would be better to simply use a single implementation - drop `washtower_common` for now, just include it directly in BDH_D39301_KR - and alias the other device ID to this implementation in `ha_bridge`.

bridge/purifier-history-store.ts
```diff
@@ -0,0 +1,323 @@
+import { readFileSync, readdirSync, renameSync, writeFileSync } from 'node:fs'
```
anszom, line 1: I would very much like to remove the history module. See the comments in the purifier device definition.

cloud/devices/FAKPK21021.ts
```diff
@@ -0,0 +1,11 @@
+import { Device as Thinq2Device } from '../thinq2/device'
```
anszom, line 1: See comments for the other washtower

cloud/devices/HWWA9K_F2.ts
```diff
@@ -0,0 +1,340 @@
...
+            }),
+        )
+    }
+
+    /*
+     * The base class strips the AA/length prefix before handing the body on, but every offset
+     * above is stated against the whole frame, the way the probes recorded them, so the frame is
+     * taken here unstripped.
+     */
+    processData(buf: Buffer) {
```
anszom, line 264: See the purifier comments - please use processAABB, there is no good reason to circumvent it.

cloud/devices/HWWA9K_F2.ts
```diff
@@ -0,0 +1,340 @@
...
+     * `f0 24` plus the three-byte body — which reproduces the captured frames byte for byte.
+     */
+    private setField(type: number, value: number) {
+        this.send(Buffer.from([...SET_STATE, type, 0x01, value]))
+    }
+
+    setProperty(prop: string, mqttValue: string) {
+        const write = (type: number, table: Record<number, string>, what: string) => {
+            const code = Object.entries(table).find(([, label]) => label === mqttValue)?.[0]
+            if (code === undefined) return console.warn(`Stick Vacuum: Unknown ${what} '${mqttValue}'`)
```
anszom, line 318: use the logging utility

cloud/devices/ST_B_E4H01Y_APL.ts
```diff
@@ -0,0 +1,828 @@
...
+import { allowExtendedType } from '@/util/casting'
+import HADevice from './base'
+import AABBDevice from './aabb_device'
+
+/*
+ * LG Styler (KR Baron Plus, S5BBP), ThinQ model ST_B_E4H01Y_APL, deviceType 203.
+ *
+ * WHERE THE BYTE OFFSETS COME FROM
+ * --------------------------------
+ * This model's modelJSON has no tlv_* labels (unlike the clip/TLV humidifier and
```
anszom, line 13: Please shorten this.

cloud/devices/ST_B_E4H01Y_APL.ts
```diff
@@ -0,0 +1,828 @@
...
+                },
+            }),
+        )
+    }
+
+    /*
+     * Offsets above are stated against the whole frame, the way the probes recorded them, so
+     * the frame is read unstripped rather than through the base class's body view.
+     */
+    processData(buf: Buffer) {
```
anszom, line 622: processAABB

cloud/devices/ST_B_E4H01Y_APL.ts
```diff
@@ -0,0 +1,828 @@
...
+                // whole and half hours, which is LG's own constraint, not ours.
+                const m = HALF_HOUR.exec(mqttValue.trim())
+                if (!m) return console.warn(`Styler: Night-care time format error '${mqttValue}'`)
+                const sub = prop === 'night_care_start' ? UP.nightCareStart : UP.nightCareEnd
+                // Hour and minute travel in one frame, exactly as LG's own setter sent them.
+                this.setUpCenter(sub, Number(m[1]), Number(m[2]))
+                return this.echo(prop, mqttValue.trim())
+            }
+            default:
+                console.warn(`Styler: Item does not support writing ${prop}`)
```
anszom, line 825: use the logging utility

cloud/devices/washtower_common.ts
```diff
@@ -0,0 +1,3322 @@
...
+    '56': 'Full-load Dry',
+    '57': 'Room Dehumidify',
+    '58': 'Turbo Dry',
+    '114': 'Wrinkle Ease Dry',
+    '255': 'Downloaded course',
+}
+
+/* LG official maps: FAKPK21021 WASHER_PROTOCOL_EX 8.3 / WASHER_CONVERT_EX 2.3;
+ * BDH_D39301_KR DRYER_PROTOCOL_EX 4.1 / DRYER_CONVERT_EX 7.8. */
+export const WASHER_MODEL: WashTowerModel = {
```
anszom, line 272: Lots of keywords that have no meaning if you don't know the context.

How was this table created?

anszom (commented):
I'll submit partial feedback for now. I've tried to review this in one go but it would take much more time than I expected.

After you go through the partial comments, please split this into one PR per device-class (so, one for humidifers, one for washtowers, etc).

anszom:
I also should have stated this at the beginning: Thank you! I appreciate the effort, this will significantly expand our support base. I apologize if my initial comments sounded dismissive - I got a bit overwhelmed with the scope of the review.

tunarabuuu:
Thanks — this is exactly the kind of review I was hoping for. I'll do the split: one PR per device-class, and go through the comments on each as I carve them out. Concretely I'm planning:

- dehumidifiers (DHUM_056905 + 231006, on the TLV path)
- humidifier (HUM_056905)
- styler
- stick vacuum
- water purifier
- washtower (I'll fold `washtower_common` into a single class as you suggested, since the two are identical today)

On the recurring points, so you don't have to repeat them per PR:
- **processAABB** — agreed, I'll route the AABB devices through it instead of the custom decode.
- **logging utility** — will switch the stray `console.*` calls over.
- **long header comments** — will trim them to what's needed to read the code.
- **purifier history / stateless** — makes sense; I'll drop the history store and see how far a stateless purifier gets. (I keep the persistent version in my own add-on, so no loss on my end.)
- **time utils / KST** — I'll move the formatting into `util/` and check whether anything is actually KST-bound (I don't think it needs to be).

I'll close this PR once the per-device ones are up, and link them here. Starting with the dehumidifiers.

tunarabuuu:
Split complete — closing this in favour of the six per-device-class PRs, each rebased on master with your review comments addressed:

- #113 — dehumidifiers (DHUM_056905 + 231006)
- #114 — humidifier (HUM_056905)
- #115 — styler (processAABB, logging util, trimmed comment)
- #116 — stick vacuum (processAABB, logging util, dropped a machine-local path)
- #117 — washtower washer + dryer (dropped `washtower_common` into a single file)
- #118 — water purifier (rewritten stateless: no history module, deltas → `total_increasing`, processAABB, KST/time logic gone)

The `HumidifierComponent` type and the `0xA7` TLV acceptance ride along in #113/#114 and are originally from @BluSyn's #64. Thanks again for the thorough review.

3735943886:
I would like to integrate your `ac_common` and `cst_170004_ww` into my work. Please let me know if you are okay with this. Thank you.

tunarabuuu:
@3735943886 Yes, absolutely — please go ahead, and thank you for asking. I'd be glad to see `ac_common` and `CST_170004_WW` land as part of your #105 work; that's exactly where they belong.

To make it easy to pull from, I've pushed the code to a reference branch on my fork: **tunarabuuu/rethink@`share-ac-common`** (https://github.com/tunarabuuu/rethink/tree/share-ac-common). It has the `ac_common` base, the `CST_170004_WW` wall/built-in IDU driver (deviceType 401), and the two small things they lean on — the reservation store and the `modes`/`preset_modes` climate discovery fields. It type-checks on current master.

Two notes for integration:
- Labels there are still Korean (that branch predates the English pass I did on the other drivers); the enum *maps* are the same protocol either way, and translate the same way. Happy to push an English-labelled version if that's easier for you.
- The `CST_170004_WW` test is from before anszom's multi-unit MQTT change merged to master, so it'll need adapting to the new HA mock — nothing about the driver logic, just the harness.

Every enum in `CST_170004_WW` was swept end to end on a live unit (single-byte injections read back through `lge_thinq`), and I still have those capture notes — say the word if any of the mappings would help. Whatever shape ac_common ends up in for the 401 family, I'm happy to rebase my wall-AC driver onto it.

PR #113, opened 2026-08-02
Add LG dehumidifiers DHUM_056905_WW and DHUM_231006_WW

tunarabuuu:
First of the per-device-class splits from #111.

Two ThinQ2 dehumidifiers on the existing TLV path (deviceType 403), each verified against the LG cloud's own decode of the same unit while bridged:

- **DHUM_056905_WW** — the base driver.
- **DHUM_231006_WW** (BEKEN_BK7234) — shares the 056905 TLV map, overrides only the mode/fan enums (it uses codes 85/86 where 056905 uses 17/21) and sends fan writes as bare `0x1fa` without the per-mode table.

Shared infrastructure this needs:
- `TLVDevice`'s packet gate accepts `0xA7` in byte 6 as well as `0x87` — the dehumidifier family frames with `0xA7`. This two-line change and the `HumidifierComponent` discovery type both originate in @BluSyn's #64; if #64 lands first I'll rebase on top of it.

### DIFF
--- a/cloud/devices/DHUM_056905_WW.ts
+++ b/cloud/devices/DHUM_056905_WW.ts
@@ -1,428 +1,57 @@
-import TLVDevice from './tlv_device'
-import { Device as Thinq2Device } from '../thinq2/device'
-import { DeviceDiscovery, type Connection, type HumidifierComponent } from '../homeassistant'
-import { type Metadata } from '../thinq'
-import { allowExtendedType } from '@/util/casting'
+import DhumDevice, { defineDhumProfile } from './dhum_common'
 import * as TLV from '@/util/tlv'
-import HADevice from './base'
 
-/**
- * TLV tags present in capability (0xA7/0x01) packets — store during the caps query phase
- * but do not publish as entity values. After caps are received, 0x336 is handled as
- * current humidity via its field definition.
- */
-const CAPS_ONLY_TAGS = new Set([0x2d5, 0x2d6, 0x336, 0x2e5, 0x2e6, 0x2da])
-
-/** Observed on bucket-empty notify when the tank is reinstalled (0x2b1=256, 0x2b2=0). */
-const BUCKET_EMPTIED_EVENT = 256
-
-/** Entering these modes resets fan speed to low (high remains user-selectable). */
-const SILENT_MODES = new Set([2, 19])
-
-const HA_MODES = ['Smart', 'Jet', 'Silent', 'Spot', 'Laundry'] as const
-
-const CLIP_TO_HA_MODE: Record<number, string> = {
-    0: 'Smart',
-    1: 'Jet',
-    2: 'Silent',
-    4: 'Spot',
-    5: 'Laundry',
-    17: 'Smart',
-    18: 'Jet',
-    19: 'Silent',
-    20: 'Spot',
-    21: 'Laundry',
-}
-
-const HA_TO_CLIP_MODE: Record<string, number> = {
-    Smart: 17,
-    Jet: 18,
-    Silent: 19,
-    Spot: 20,
-    Laundry: 21,
-}
-
-/**
- * Per-mode fan capability table resent with every fan-speed write.
- *
- * The panel/device does not treat fan speed as a single global value alone: a write of
- * 0x1fa must be accompanied by a repeating 3-tuple capability declaration:
- *   0x2d7 = operating-mode id
- *   0x2d8 = 0 (always observed)
- *   0x2d9 = fan speed that mode should run at (2 = low, 6 = high)
- *
- * Most modes accept the requested fan speed. Laundry is fixed at high — that is a
- * property of the mode, not a special-case in the write path. Mode 22 appears in the
- * on-wire table on observed units but is not exposed in the ThinQ app / HA modes list.
- *
- * This is a capability declaration, not device state: it must not be stored in
- * raw_clip_state (those tags are not globally unique — they repeat once per mode row).
- */
-type ModeFanCap = {
-    mode: number
-    /** When set, 0x2d9 is always this value; otherwise it follows the requested fan. */
-    fixedFan?: 2 | 6
-}
-
-const MODE_FAN_CAPS: readonly ModeFanCap[] = [
-    { mode: 17 }, // Smart
-    { mode: 18 }, // Jet
-    { mode: 20 }, // Spot
-    { mode: 21, fixedFan: 6 }, // Laundry — high fan only
-    { mode: 22 }, // present on-wire; not exposed in app
-]
-
-function normalizeHaMode(val: string): string {
-    return val.charAt(0).toUpperCase() + val.slice(1).toLowerCase()
-}
-
-/**
- * LG Dehumidifier DHUM_056905_WW (e.g. models using 056905 platform, deviceType 403)
- */
-export default class Device extends TLVDevice {
-    powerStatePrev?: boolean
-    modePrev?: string
-    modeClipPrev?: number
-    initialValuesReceived = false
-    /** Last bucket-full state published to HA (retained). */
-    bucketFullHaState?: boolean
-
-    constructor(HA: Connection, thinq: Thinq2Device, meta: Metadata) {
-        super(HA, thinq)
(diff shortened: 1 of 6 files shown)


anszom:
Can you rebase this to master? I've merged #64 first so there are some conflicts. Also, the comments mention Korean labels, but all labels are in English now.

tunarabuuu:
Rebased onto master now that #64 is in — the conflicts are gone and it is one commit again.

Rather than resolve my old `DHUM_056905_WW.ts` against the merged one, I dropped mine and built on yours. Your driver stays as merged; the only change to it is that the model-specific tables (mode enum, fan enum, the per-mode fan-capability rows, the silent-mode set) move onto a `static modeTables` so a sibling model on the same platform can override just those. Everything my old branch carried on top of #64 — the held-back optional switches, the extra LG-declared entities, the `Spot`→`Intensive` rename, the 0x1fd sensor — is out; they were changes to a driver you had just taken, and none of them is needed for the second model.

So this is now **DHUM_231006_WW only**: same TLV map, different mode codes (19/20/85/86 vs 17/18/20/21/22) and five fan speeds instead of two, plus the base-class generalization those need, a README line and its tests.

You are right about the comments — the Korean-label notes were left over from before the English pass. `DHUM_231006_WW.ts`'s header and the test names no longer claim it; the labels there are the English ones the file actually publishes, and the fan options are lower-case `low`/`high`/… to match what your driver already exposes.

`tlv_device.ts` needed nothing: the 0xA7 acceptance is already on master.

414 tests passing (409 on master + 5 here), `tsc --noEmit` and `prettier --check` clean.

One caveat on the README line: I could not find this unit's marketing model number in anything I have captured — LG's own modelJSON for the platform just says `Full Model` — so it is listed by its ThinQ id, the way `F3L7CYK5W_US_WIFI` and friends are. Happy to replace it with the sales model as soon as I can read it off the appliance.

cloud/devices/DHUM_056905_WW.ts
```diff
@@ -72,14 +72,57 @@ const MODE_FAN_CAPS: readonly ModeFanCap[] = [
...
+        haModes: HA_MODES,
+        clipToHa: CLIP_TO_HA_MODE,
+        haToClip: HA_TO_CLIP_MODE,
+        silent: SILENT_MODES,
+        fanOptions: ['low', 'high'],
+        fanToHa: { 2: 'low', 6: 'high' },
+        fanToClip: { low: 2, high: 6 },
+        lowFan: 'low',
+        lowFanClip: 2,
+        fanClipValues: new Set([2, 6]),
```
anszom: there is a lot of duplication here, easy to get subtly wrong. Maybe it would be better to have a function that would accept a simplified mode table and return an expanded one with the inverted tables. For example all the fan options could be derived from the current value of `fanToClip`, either here, or where it is used.

cloud/devices/DHUM_056905_WW.ts
```diff
@@ -72,14 +72,57 @@ const MODE_FAN_CAPS: readonly ModeFanCap[] = [
...
+    fanToClip: Record<string, number>
+    /** Label and code of the lowest fan speed. */
+    lowFan: string
+    lowFanClip: number
+    /** Codes this model accepts for a fan write; anything else is dropped. */
+    fanClipValues: ReadonlySet<number>
+    /** Capability rows to send alongside 0x1fa; empty for a model with no capture of them. */
+    modeFanCaps: readonly ModeFanCap[]
+}
+
```
anszom: With this refactor, it would be reasonable to move most of this file to dhum_common.ts and base both dehumidifiers on that common base.

But see the comment about `modeFanCaps` below.

cloud/devices/DHUM_231006_WW.ts
```diff
@@ -0,0 +1,63 @@
...
+        lowFanClip: 2,
+
+        // Five speeds, not the 056905's two.
+        fanClipValues: new Set([2, 4, 6, 7, 8]),
+
+        // The 056905 rewrites its whole per-mode fan-memory table on every fan change. No
+        // capture shows this model doing that, and its mode codes are different anyway, so
+        // emitting that table here would be inventing traffic. Send 0x1fa alone until a
+        // panel capture says otherwise.
+        modeFanCaps: [],
```
anszom: If I understand this correctly, this will cause an empty ([1,1,2,1,1]) packet to be emitted for each fan mode change. 

I think a better approach would be to keep an empty `sendFanSpeedTlvs` in the base class and leave the fan table rewrite code in 056905's implementation only.

tests/cloud/devices/DHUM_231006_WW.test.ts
```diff
@@ -0,0 +1,106 @@
...
+import type { Metadata } from '@/cloud/thinq'
+import * as TLV from '@/util/tlv'
+import { MockHAConnection, MockThinq2Device, buf } from '@/tests/helpers/mocks'
+import { enableMockTimers, tickMockTimers } from '@/tests/helpers/timers'
+
+const DEVICE_ID = 'test-id'
+const MODEL_ID = 'DHUM_231006_WW'
+const META: Metadata = { modelId: MODEL_ID, modelName: 'TEST DHUM', swVersion: '1.0' }
+
+// Real 056905-platform captures; this model shares the initialization path.
```
anszom: Either this is a copy-paste bug, or you're testing on the wrong captures here.

cloud/devices/DHUM_231006_WW.ts
```diff
@@ -0,0 +1,63 @@
...
+        haToClip: {
+            'Smart Plus': 86,
+            Silent: 19,
+            Intensive: 20,
+            Quick: 85,
+        },
+        // Entering Silent drops the fan to its lowest setting, as the 056905 does.
+        silent: new Set([19]),
+
+        fanOptions: ['auto', 'low', 'mid', 'high', 'turbo'],
```
anszom: 'medium' would be a better name, unless 'mid' is clearly the official one.

anszom (changes requested):
This looks better, but I'd still like the refactor to be shaped a bit differently, see my comments below.

anszom:
> One caveat on the README line: I could not find this unit's marketing model number in anything I have captured

This is quite common, I guess that the same electronics may be packaged in various market models and they will truly have no information on the marketing model name. Reading the sticker is the only sure way.

tunarabuuu:
Rebased this PR onto current `master` and addressed the review feedback with a model-specific, capture-backed implementation.

Key changes:
- factored the shared 056905/231006 dehumidifier behavior into a common profile-driven base instead of duplicating the full driver;
- kept the 231006 model surface limited to entities confirmed by its own real capture (no inherited `bucket_full`);
- used the captured 231006 mode/fan tables and captured plain `0x1fa` writes, with exact `unknown` fallbacks and rejected unknown commands;
- added the actual `DHUM_231006_WW` response capture to its tests instead of reusing a 056905 fixture;
- fixed initial startup for this model: its observed first response combines live state with `0x2d7/0x2d8/0x2d9` capability rows and has no `0x2da`, so the driver now recognizes that real packet shape and completes both capability and values initialization without a test-only timer bypass.

Fresh verification on final commit `1419bf6`:
- full test suite: 450/450 passed;
- targeted 231006 tests: 6/6 passed;
- build and build TypeScript check passed;
- Prettier and `git diff --check` passed;
- GitHub CI: Node 20, Node 24, and unlabel checks all passed.

cloud/devices/dhum_common.ts
```diff
@@ -0,0 +1,424 @@
...
+                t === 0x1f7 ||
+                t === 0x1f9 ||
+                t === 0x1fa ||
+                t === 0x21b ||
+                t === 0x21e ||
+                t === 0x2b2 ||
+                t === 0x253 ||
+                t === 0x2a2 ||
+                t === 0x336 ||
+                t === 0x360,
```
maciejsszmigiero, line 414: So if the device reports that the current humidity has changed while being probed this single tag change notification will be regarded as full values query response?

That's _not_ how this should work, see `isValuesResponse()` in `RAC_056905_WW.ts` how this should be done instead.

PR #115, opened 2026-08-02
Add LG Styler ST_B_E4H01Y_APL

tunarabuuu:
Third of the per-device-class splits from #111.

**ST_B_E4H01Y_APL** — an AA..BB ThinQ2 Styler / garment-care unit (deviceType 203). The byte layout was recovered by injecting single-byte-changed state frames through the management API and reading the LG cloud's own decode back, one offset per observation (three of the bytes are bitfields, split bit by bit).

Addressing your review notes from #111:
- **processAABB** — reads now go through the base class's `processAABB` instead of a custom `processData`; the frame arrives with its leading AA/FF and trailing CRC/BB stripped, so the record offsets are simply two less than the whole-frame positions.
- **logging utility** — the stray `console.warn` calls now use `log(...)`.
- **header comment** — trimmed to what's needed to read the code.

### DIFF
--- a/cloud/devices/ST_B_E4H01Y_APL.ts
+++ b/cloud/devices/ST_B_E4H01Y_APL.ts
@@ -1,0 +1,736 @@
+import { Device as Thinq2Device } from '../thinq2/device'
+import log from '@/util/logging'
+import { DeviceDiscovery, type Connection } from '../homeassistant'
+import { type Metadata } from '../thinq'
+import { allowExtendedType } from '@/util/casting'
+import HADevice from './base'
+import AABBDevice from './aabb_device'
+
+/*
+ * LG Styler (S5BBP), ThinQ model ST_B_E4H01Y_APL, deviceType 203.
+ *
+ * The byte layout was recovered by injecting single-byte-changed state frames through the
+ * management API and reading LG cloud's decode back, so each offset rests on a direct
+ * observation. Three of the bytes are bitfields, split bit by bit.
+ *
+ * READ. `aa ff 31 0a 00 | <len16 LE> | <seq16> | 00 01 | 00 <tag> | <record bytes, BE16> |
+ * …40-byte records… | <crc16 XMODEM> bb`. Offsets into the whole frame, single-record form:
+ *
+ *   17 state              18 remainTimeHour     19 remainTimeMinute
+ *   20 initialTimeHour    21 initialTimeMinute  22 course
+ *   23 error              24 preState
+ *   29 reserveTimeHour    30 reserveTimeMinute
+ *   31 flags  0x01 childLock  0x02 nightDry  0x04 initialBit  0x08 remoteStart
+ *   34..35 energyMonitoring, big-endian 16-bit — not published, the unit is undeclared
+ *   37 smartCourse
+ *   40 currentDownloadCourseCount   41..44 currentDownloadCourse1..4
+ *   45 buzzer
+ *   46 flags  0x01 applyRemoteMaintain  0x02 applyBuzzer  0x04 remoteMaintain
+ *   47 endMelody          48 internalLightingTime — not published, no declared value list
+ *   49 flags  0x01 isLastCourse  0x02 smartCareFineDust  0x04 smartCareHumidity
+ *             0x08 smartCareNightCare  0x10 smartPairing  0x20 currentTimeDisplay
+ *   50..53 nightCareStartTime / nightCareEndTime, hour then minute
+ *   54 TCLCount
+ *
+ * That covers all 38 `styler.*` keys LG publishes for this unit. Byte 31 bit 0x20 and byte 49
+ * bits 0x40/0x80 move no field and are left alone.
+ *
+ * The record count varies, so it is counted rather than assumed: at rest the appliance sends
+ * one 40-byte record, and while it is doing anything it sends two — previous, then current —
+ * as the water purifier in this repo does. Header 15 bytes, trailer 3, and what is between
+ * must be a whole number of records, cross-checked against the record-bytes field at 13..14.
+ * That also rejects the frames which carry no state: the 112/113-byte downloadable-course name
+ * lists, the 35-byte `aa ff 31 0a 00 23 00 …` (every byte probed with 0xff, nothing moved) and
+ * the `aa 07 31 c3 02` heartbeat.
+ *
+ * WRITE. Settings go out as
+ *
+ *   aa <len> f0 24 <controlDataType> <valueLength> [<controlDataType_sub>] <value…> <ck> bb
+ *
+ *   01 POWERON/POWEROFF          len 1   off 0, on 1
+ *   04 PAUSE                     len 1   0
+ *   10 REMOTE_MAINTAIN           len 1   off 0, on 1
+ *   13 UPCENTER, then a sub id and the value:
+ *      01 ENDMELODY              len 1   melody index, 0..11
+ *      03 UPBUZZER               len 1   buzzer level, 0..4
+ *      04 SC_FINEDUST            len 1   off 0, on 1
+ *      05 SC_HUMIDITY            len 1   off 0, on 1
+ *      06 SC_NIGHTCARE           len 1   off 0, on 1
+ *      07 SC_NIGHTCARE_START     len 2   hour, minute
+ *      08 SC_NIGHTCARE_END       len 2   hour, minute
+ *      09 CTRL_IS_LAST_COURSE    len 1   off 0, on 1
+ *
+ * and course control as
+ *
+ *   start  aa 34 f0 26 | <46-byte course block> | ck bb
+ *   resume aa 33 f0 26 | <course id> <44 zero bytes> | ck bb
+ *
+ * `controlDataType`, `controlDataValueLength` and `controlDataType_sub` are modelJSON's own
+ * `ControlWifi` vocabulary, and every value is the code the read map already uses for that
+ * field. The frames came from LG's own cloud (`dataops/v1/s2p/backend/convert/control`), which
+ * wants `{"setOnOff":{"value":"on"}}` rather than the aircon's `{"onOff":"on"}`, and answers
+ * CL-0005 while the appliance is in no state to take the command. REMOTE_MAINTAIN is the one
+ * exception: LG's converter only ever emits its OFF frame, so the ON frame was built from the
+ * same shape, sent to the appliance directly, acknowledged, and read back on byte 46 bit 0x04.
+ *
+ * The course block is `<id> 01 00 04 00 00 00 00 00` followed by modelJSON's
+ * `Course[id].function` defaults in declaration order, with `TimeDry` folded into bit 0x80 of
+ * the `PreSteam1_Time` byte. Standard (id 1) and Quick (id 3) were captured off LG's converter
+ * and the suite checks the frames this driver builds for them byte for byte.
+ *
+ * `cycles setCurrentCycle` does not convert for this model, so course select is a local choice
+ * this driver remembers and Start course is what sends it — the order LG's own app uses too.
+ */
+const HEADER_LEN = 13
+const TRAILER_LEN = 1
+const RECORD_LEN = 40
+/** buf[13..14], big-endian: how many record bytes the frame says it carries. */
+const RECORD_BYTES_OFF = 11
+const STATE_TAG = 0x31
+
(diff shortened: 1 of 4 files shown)


cloud/devices/ST_B_E4H01Y_APL.ts
```diff
@@ -0,0 +1,811 @@
...
+    3: '82645a00000003005a00000001b4000eb40000000000000000000000000000000000000000', // Styling Quick
+    5: '02645a00000007005a05c85a01c80031b40000000000000000000000000000000000000000', // Styling Intensive
+    6: '82000000000004000000000001000014780000000000000000000000000000000000000000', // Wool/Knit
+    7: '82645a00000004005a04b45a01b40017780000000000000000000000000000000000000000', // Suit/Coat
+    11: '0200002d000006000008000003000023780000000000000000000000000000000000000000', // Sterilize Standard
+    15: '0000000000000000000000000000005a780000000000000000000000000000000000000000', // Auto Dry
+    17: '8000000000000000000000000000001e780000000000000000000000000000000000000000', // Timed Dry 30
+    18: '8000000000000000000000000000003c780000000000000000000000000000000000000000', // Timed Dry 60
+    19: '8000000000000000000000000000005a780000000000000000000000000000000000000000', // Timed Dry 90
+    20: '8000000000000000000000000000003c780000000000000000000000000000000000000000', // Timed Dry 120
```
anszom, line 169: This is identical to Timed Dry 60. Is this correct?

cloud/devices/ST_B_E4H01Y_APL.ts
```diff
@@ -0,0 +1,811 @@
...
+                        icon: 'mdi:sleep',
+                        entity_category: 'config',
+                    }),
+                    night_care_start: halfHourClock('night_care_start', 'Night-care start time'),
+                    night_care_end: halfHourClock('night_care_end', 'Night-care end time'),
+                    // Replaced by the two settable halves above. Removing a component from the
+                    // config is not enough to retire the entity Home Assistant already made —
+                    // a config carrying nothing but its platform is what says "this is gone",
+                    // and it has to carry that, or HA rejects the whole device payload.
+                    night_care_window: { platform: 'sensor' } as ComponentInfo,
```
anszom: That entity was never exposed by a released version of rethink, so we don't need to include code to withdraw it.

cloud/devices/ST_B_E4H01Y_APL.ts
```diff
@@ -0,0 +1,811 @@
...
+                    course_select: choice('course_select', 'Course select', COURSE_OPTIONS, {
+                        icon: 'mdi:playlist-edit',
+                    }),
+                    start_course: press('start_course', 'Start course', 'mdi:play-circle-outline'),
+                    pause_course: press('pause_course', 'Pause', 'mdi:pause-circle-outline'),
+                    resume_course: press('resume_course', 'Resume', 'mdi:play-pause'),
+                    smart_course: sensor('smart_course', 'Smart course', { icon: 'mdi:playlist-star' }),
+                    remaining: minutes('remaining', 'Time remaining', { icon: 'mdi:timer-sand' }),
+                    total: minutes('total', 'Total time', { icon: 'mdi:timer-outline' }),
+                    reserved_at: sensor('reserved_at', 'Reserved time', { icon: 'mdi:calendar-clock' }),
```
anszom: Please follow the naming conventions of other devices (remaining_time, reserve_time)

cloud/devices/ST_B_E4H01Y_APL.ts
```diff
@@ -0,0 +1,811 @@
...
+            entity_category: 'config',
+            ...extra,
+        })
+
+        this.setConfig(
+            allowExtendedType({
+                ...HADevice.config(meta, { name: 'LG Styler' }),
+                components: {
+                    power: toggle('power', 'Power', { icon: 'mdi:power' }),
+                    status: sensor('status', 'State', { icon: 'mdi:hanger' }),
```
anszom: Other devices use the 'Status' label

cloud/devices/ST_B_E4H01Y_APL.ts
```diff
@@ -0,0 +1,811 @@
+import { Device as Thinq2Device } from '../thinq2/device'
+import log from '@/util/logging'
+import { DeviceDiscovery, type ComponentInfo, type Connection } from '../homeassistant'
+import { type Metadata } from '../thinq'
+import { allowExtendedType } from '@/util/casting'
+import HADevice from './base'
+import AABBDevice from './aabb_device'
+
+/*
```
anszom, line 9: This header is overly verbose. Keep only the most important protocol details, definitely drop mentions of older revisions that we have never seen.

cloud/devices/ST_B_E4H01Y_APL.ts
```diff
@@ -0,0 +1,811 @@
...
+ * which lists names but no codes; the code is the position in that list. Three points were
+ * observed on the wire and all three agree — byte 17 = 1 read back INITIAL (index 1),
+ * byte 24 = 2 read back RUNNING (index 2), byte 23 = 1 read back ERROR_TE1 (index 1) — so
+ * the remaining entries follow the same declaration order.
+ *
+ * `styler.course` and `styler.smartCourse` are NOT positional: the byte is LG's course id
+ * straight out of the `Course` / `SmartCourse` tables. Confirmed twice — byte 22 = 1 read
+ * back STANDARD (id 1), and the untouched byte 41 = 0x4e = 78 read back FUR_LEATHER (id 78).
+ *
+ * Korean text is from this model's own ko-KR packs: the product pack for states and melodies,
```
anszom: various references to korean text are outdated

tests/cloud/devices/ST_B_E4H01Y_APL.test.ts
```diff
@@ -0,0 +1,657 @@
...
+    const ha = new MockHAConnection()
+    const thinq = new MockThinq2Device(DEVICE_ID, META)
+    const dev = new DUT(ha.asConnection(), thinq, META)
+    return { ha, thinq, dev }
+}
+
+describe(MODEL_ID, () => {
+    /*
+     * Writable exactly where a command frame was captured off LG's own capability API and seen to
+     * take (CL-0000). Deliberately absent: `remoteControlStatus setRemoteControlType`, which
```
anszom: The fields claimed to  be absent do appear to be present...

tests/cloud/devices/ST_B_E4H01Y_APL.test.ts
```diff
@@ -0,0 +1,657 @@
...
+        'start_course',
+        'pause_course',
+        'resume_course',
+    ])
+
+    test('exactly the fields with a captured command frame are writable', () => {
+        const { ha } = makeDevice()
+        const components = ha.devices[DEVICE_ID].config!.components as Record<string, Record<string, unknown>>
+        for (const [name, comp] of Object.entries(components)) {
+            if (comp.unique_id === undefined) continue // a removal, not an entity
```
anszom: This line should not be needed if we remove the night_care_window reference entirely.

tests/cloud/devices/ST_B_E4H01Y_APL.test.ts
```diff
@@ -0,0 +1,657 @@
...
+        assert.equal(ha.devices[DEVICE_ID].properties.night_care_end, '06:00')
+    })
+
+    test('the retired night-care summary is published as a removal', () => {
+        const { ha } = makeDevice()
+        const components = ha.devices[DEVICE_ID].config!.components as Record<string, Record<string, unknown>>
+        // Its two settable halves replace it, but dropping it from the config would leave the old
+        // entity behind, frozen at whatever it last read.
+        assert.deepEqual(components.night_care_window, { platform: 'sensor' })
+    })
```
anszom: This should not be needed if we remove the night_care_window reference entirely.

cloud/devices/ST_B_E4H01Y_APL.ts
```diff
@@ -0,0 +1,811 @@
...
+                    tcl_count: sensor('tcl_count', 'Sterilize-tank clean alert count', {
+                        icon: 'mdi:counter',
+                        entity_category: 'diagnostic',
+                        // Resets when the maintenance course runs, so total_increasing would
+                        // wrongly accumulate across the reset — no state_class.
+                    }),
+                    energy: sensor('energy', 'Energy monitoring', {
+                        icon: 'mdi:flash',
+                        entity_category: 'diagnostic',
+                    }),
```
anszom: Is this in a useful real-world unit?

cloud/devices/ST_B_E4H01Y_APL.ts
```diff
@@ -0,0 +1,811 @@
...
+                        icon: 'mdi:history',
+                        entity_category: 'diagnostic',
+                    }),
+                    internal_light: sensor('internal_light', 'Interior light setting', {
+                        icon: 'mdi:lightbulb-outline',
+                        entity_category: 'diagnostic',
+                        // LG's snapshot exposes styler.internalLightingTime but this model's
+                        // modelJSON does not declare its value list, so the raw code is published
+                        // rather than an invented label. 0 read back NO_SETTING, 1 read LIGHTING_TIME_0.
+                    }),
```
anszom: What is the meaning of this entity? If it has only two values, would a binary_sensor be a better choice?

anszom:
Also please add a README entry just as for the other devices.

tunarabuuu:
All addressed, one commit, rebased on master.

- **Header too verbose (:9)** — cut from 105 lines to about 70, and everything that narrated this driver's own history is gone ("the earlier note said control was refused", "which the first pass got wrong", the retelling of what an older revision published). What is left is the read map, the frame/record shape, the write vocabulary and the course-block rule — the parts you need to read the code. The same treatment went to the `styler.state` / `styler.error` comment blocks and to one comment in the test file.

- **Outdated Korean references (:264)** — gone. Where the source of a label matters the comment now says it is an English translation of LG's ko-KR pack, which is what the tables actually hold.

- **`Timed Dry 120` identical to `Timed Dry 60` (:218)** — it is correct, and it surprised me too, so I checked it against the model file rather than my transcription. LG's modelJSON gives course 20 a `Drying1_Time` default of **60**, exactly as course 18 does; I regenerated the whole table from `Course[id].function` with the documented rule and every one of the 21 entries came back byte for byte identical to what is in the PR. The run length evidently comes from the course id, not from this block. There is now a comment saying so.

- **Naming conventions (:528)** — `remaining` → `remaining_time`, `total` → `initial_time`, `reserved_at` → `reserve_time`, matching `F_V8_Y___W.B_2QEUK` and friends. `reserve_time` is now a duration in minutes rather than a clock string: LG declares `Reserve_Time_H` as a 3..19 hour range, so it is a delay before the course starts, and 0 means no reservation.

- **`'Status'` label (:515)** — done.

- **`night_care_window` withdrawal (:573, tests :124 and :582)** — you are right, that entity never shipped. The tombstone component, the test for it, and the `unique_id === undefined` skip in the writability test are all gone.

- **Energy monitoring in a useful unit? (:587)** — no, and I could not make it one. The two probes only establish that 34..35 is a big-endian 16-bit value; LG's `EnergyMonitoring` block for this model is a `powertable` keyed by level, which does not say what the counter counts. Rather than publish a number with no unit I dropped the entity (and its probe fixture). The offset stays documented in the header.

- **`internal_light` meaning (:598)** — dropped for the same reason. `styler.internalLightingTime` has no value list in this model's modelJSON and nothing in its language pack, and the two codes I saw (0 → `NO_SETTING`, 1 → `LIGHTING_TIME_0`) do not settle whether "time 0" means the light is off or on — so a binary sensor would have a coin-flip polarity. If a capture ever pins it down it is a two-line addition.

- **Test :99** — the comment was stale: it claimed `remoteControlStatus` and course start/pause/resume were deliberately absent, while all four are in the writable set. It now says what the rule actually is — writable where a frame the appliance accepts was established, either captured from LG's converter or (for remote-maintain ON, which the converter never emits) built from the same shape, acknowledged, and read back on the next state frame.

442 tests passing (409 on master + 33 here), `tsc --noEmit` and `prettier --check` clean. README entry added as `S5BBP (ST_B_E4H01Y_APL)`.

anszom:
I've made some small adjustments for consistency with other devices & merged. Closing PR manually. Thanks for the contribution!

PR #117, opened 2026-08-02
Add LG Korean WashTower washer and dryer

tunarabuuu:
Fifth of the per-device-class splits from #111.

Adds the Korean WashTower washer **FAKPK21021** (deviceType 221) and dryer **BDH_D39301_KR** (222).

Review follow-up:
- replaced automatic publication of 141 washer and 84 dryer protocol fields with a hand-picked HA surface: state, course, error, reserve/remaining/initial durations, locks, remote flags, washer door, verified settings, and existing guarded controls;
- added HA duration/enum/door metadata and the exact lowercase `unknown` fallback;
- runtime decoding now tracks only fields needed for the public surface or controls; the large official field maps remain internal codec/control evidence and no longer create raw diagnostic entities;
- tests pin the exact public property lists and state decoded from real washer 0xEB, dryer 0xEB, and dryer 0xEC captures; table-vs-itself bitfield/manifest tests were removed;
- rebased onto current `master` and retained all current device registrations.

**Relationship to `WTL_FXU_BDV_NA_01`:** the useful HA semantics overlap, and this implementation now uses the same concepts and topic names where applicable. The wire protocols do not: WTL uses its combined 0xD0/0x71 handler, while these Korean split devices use distinct `AA FF` extended records with separate 0xEB/0xEC washer/dryer tags and model-specific control payloads. Reusing or aliasing the WTL codec would therefore decode the wrong framing and offsets.

Fresh verification on `3b1942c`: targeted WashTower tests 38/38, full suite 482/482, build/typecheck, Prettier, and diff check all pass.

### DIFF
--- a/cloud/devices/washtower.ts
+++ b/cloud/devices/washtower.ts
@@ -1,0 +1,3351 @@
+import { Device as Thinq2Device } from '../thinq2/device'
+import { type Connection, type DeviceDiscovery } from '../homeassistant'
+import { type Metadata } from '../thinq'
+import { allowExtendedType } from '@/util/casting'
+import crc16 from '@/util/crc16'
+import HADevice from './base'
+import log from '@/util/logging'
+
+/*
+ * LG Korean WashTower washer (FAKPK21021, deviceType 221) and dryer (BDH_D39301_KR, 222).
+ * These speak a dedicated 0xEB/0xEC extended-frame protocol, not the AA..BB or clip/TLV forms.
+ *
+ * The field maps below (offset, length, and the raw->label enums) come from LG's own protocol
+ * and convert maps for each model (named at each MODEL constant) cross-checked against the
+ * appliance's capability dump; a field is included only where both agree. Each MODEL is a table
+ * of what byte holds what and how its raw values decode — the washer and dryer share all the
+ * decode/control code below and differ only in these two tables.
+ */
+type Field = {
+    key: string
+    offset: number
+    length: number
+    bit?: number
+    bits?: number
+    label: string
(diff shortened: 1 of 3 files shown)


anszom:
I appreciate your PR series, although this one case is problematic.

The protocol handler produces a huge number of entities in HA, none of them associated with any user-friendly metadata (classes, units, icons). This is an understandable consequence of machine-mapping the protocol definition. I have a feeling that a hand-picked set of entities, mapped individually, would be more useful.

Does this implementation have anything in common with WTL_FXU_BDV_NA_01? Claude claims that the protocols overlap significantly.

Most of the tests verify only some implementation details, not testing against actual captured packets.

Honestly, I'm not sure how to proceed with this.

tunarabuuu:
Addressed the maintainer feedback in `3b1942c`.

- HA no longer receives every machine-mapped protocol field: the driver now publishes only the hand-picked state/cycle/time/error/lock/remote/door surface plus verified settings and controls.
- Added useful HA metadata and exact `unknown` fallbacks.
- Removed implementation-only tests that checked generated tables against themselves; real 0xEB/0xEC captures now pin the public entity surface and decoded state.
- Kept only control/public fields in the runtime decode path.
- Clarified the WTL question in the PR body: HA concepts overlap, but WTL 0xD0/0x71 framing is not wire-compatible with the Korean split-device `AA FF` 0xEB/0xEC records, so sharing that codec would be incorrect.

Fresh verification: WashTower tests 38/38, full suite 482/482, build/typecheck, Prettier, and diff check passed. GitHub CI passed on Node 20 and Node 24.

PR #120, opened 2026-08-02
Add support for LG Studio Hood

B1223GS87:
LG Studio Hood Model HCED3015D AA..BB ThinQ2  (deviceType 304) "STUDIO_HOOD"
Supports a 5 speed fan control, and 2 intensity light. 
Tested and verified on real hardware with Home Assistant.  I used bridge mode from the LG App and captured data. Testing also involved a power cycle of the device, and validating Home Assistant updates when the device is controlled locally.


anszom:
Thanks for the PR. That's a curious model name, but that's on LG :)

The unit tests are currently failing, please fix this. Also, please add a README entry for this device.

B1223GS87:
It's a very generic model name, but ah well :) I wonder if other model hoods use the same name. I'll work on fixing the unit test and adding the readme this weekend. Thank you!

B1223GS87:
@anszom This should be good to go, let me know If i missed anything.

anszom:
Looks good, thanks! I've squash-merged it manually, closing the PR.

PR #121, opened 2026-08-03
Dashboard UI Improvements.

3735943886:
**Show the ThinQ name for each appliance.** The panel listed a UUID and a model id;
The bridge already reads the account's aliases for registration — this keeps the list and
hands it to the panel, refreshed on login, register and logout. Costs nothing to an
installation without an account: no row has a name, so the column is not drawn.

**Reconnect a page restored from the back/forward cache.** Back from the monitor landed
on an empty panel for five seconds: the browser closes the socket when it caches the
page, so the close handler ran on the way back in.

**Fit the device list on a phone.** Six columns, one a 36-character UUID, scrolled
sideways. Below the breakpoint each row becomes a card, with the bridge switch and
monitor button pinned right. Nothing is dropped — name and id ellipsize and carry their
full text as `title`. Auto-scroll also moves out of the "Messages" heading, which was
acting as a float container and splitting the switch across a wrap.

### DIFF
--- a/bridge/index.ts
+++ b/bridge/index.ts
@@ -93,12 +93,22 @@ class BridgedDevice {
 type BridgeEvents = {
     loggedIn: () => void
     loggedOut: () => void
+    namesChanged: () => void
     started: (id: string) => void
     stopped: (id: string) => void
 }
 
 export class Bridge extends TypedEmitter<BridgeEvents> {
     bridgedDevices = new Map<string, BridgedDevice>()
+
+    /*
+     * What the owner calls each appliance, from the ThinQ account - the same names the app shows.
+     *
+     * rethink knows a device by its id and its model, which is enough to talk to it and useless for
+     * telling four identical ceiling cassettes apart. The account already holds the answer; it was
+     * only ever read while registering a device (see registrationPlan), so it never reached the panel.
+     */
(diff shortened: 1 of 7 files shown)


anszom:
1. I've cherry-picked the first commit.
2. The second one (names) seems to rely on the "adopt existing devices" scheme you proposed (I still have to review that one, it's the most complex part). Otherwise, with the current bridge implementation, the ThinQ-side names will all be "Rethink xyzzy", and even if the user updates them, we only refresh on restart.
3. The third commit is good to merge, but it depends on the second.

3735943886:
Thanks for reviewing this.

Regarding the names (2): I expect most people will rename them. Anyone turning the bridge on is doing so to use the app, and the app is where the name needs to be readable. (I renamed all of mine too back when I used the upstream repo).

Regarding refreshing only at startup: Fair point, and I'd like to fix it in this PR if you don't mind. My preference is to trigger a refresh when the management panel is opened, or alternatively, we could add an explicit sync button with no auto-refresh, or simply drop the name feature altogether.

Let me know your thoughts.

anszom:
Sorry for the delay. I've rebased your work onto master, made some follow-up changes and merged. Closing the PR.

PR #122, opened 2026-08-03
Pr/ac_common

### LINKED ISSUE #105: Plan: common AC logic into ac_common

3735943886:
This is the `ac_common` base class that came out from the review on #101.

**`CST_570004_WW`** (ceiling cassette, deviceType 401) is the only unit confirmed on real
hardware — I own these, and every wire value was driven and read back.

**Everything else — RAC, POT, WIN, and PAC — is existing code moved onto
`ac_common`. The tests pass, but none of it is verified on hardware; I have none of those
units.** PAC's values come from @kkqq9320's captures in #105.

Corrections and "this does nothing on my unit" reports are very welcome, and I am happy to
rework or split this however you prefer.

Closes #105


anszom:

PR #123, opened 2026-08-03
Add support for F_V7_Y___W.B__QEUK/FV1409S4W

danleongjy:
Thank you for your excellent work enabling local control for LG ThinQ devices.  I have added support for a front-loading washing machine F_V7_Y___W.B__QEUK/FV1409S4W, using the handler for F_V8_Y___W_B_2QEUK as a base.  The main change I've made is to remove TurboWash and Energy sensors as these are not featured on F_V7_Y___W.B__QEUK.

### DIFF
--- a/tests/cloud/devices/F_V7_Y___W.B__QEUK.test.ts
+++ b/tests/cloud/devices/F_V7_Y___W.B__QEUK.test.ts
@@ -1,0 +1,134 @@
+import { describe, test } from 'node:test'
+import assert from 'node:assert/strict'
+import DUT from '@/cloud/devices/F_V7_Y___W.B__QEUK'
+import type { Metadata } from '@/cloud/thinq'
+import { MockHAConnection, MockThinq2Device, buf } from '@/tests/helpers/mocks'
+
+const DEVICE_ID = 'test-id'
+const MODEL_ID = 'F_V7_Y___W.B__QEUK'
+const META: Metadata = { modelId: MODEL_ID, modelName: 'F_V7_Y___W.B__QEUK', swVersion: '0.0.0' }
+
+// All hex dumps below are captured status telegrams from issue #75 (F6WV710P2S.ABLQPDG).
+// The decoder parses the second record (starting at buf[42]); the option state lives at
+// buf[50] (wash intensity), buf[53] (extra rinse count) and the buf[57] options bitfield.
+
+// Cotton, 40°C, 1600 RPM — no extra options selected.
+const SAMPLE_COTTON_STD = buf(
+    'aa5420ec002501000000000000000000000000000000000000000001006400000000000000000000000000002501033a033a0100030b04010000000000010002000001006400000400000000000000000000e0bb',
+)
+
+// Allergy Care, 60°C, Steam, 1600 RPM.
+const SAMPLE_ALLERGY_STEAM = buf(
+    'aa5420ec002501021c021c0900030504010000000000000002000001006400000500000000000000000000002501023002302d00030b06010000008000000003000001006400000500000000000000000000f2bb',
+)
+
+// Cotton, 60°C, Steam, 1600 RPM.
+const SAMPLE_COTTON_STEAM = buf(
+    'aa5420ec002501031c031c0100030b03010000000000010002000001006400000400000000000000000000002501041104110100030b0601000000800001000300000100640000040000000000000000000015bb',
+)
+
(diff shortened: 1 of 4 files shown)


Drafteed (approved):

anszom:
Thanks for the PR. I'm not sure why, but the main file `F_V7_Y___W.B__QEUK.ts` looks like was based on a type-erased javascript output - it is still typescript, but only technically. I don't think this was intended.

danleongjy:
Thanks for taking a look at this @anszom. Apologies I had generated the previous version of `F_V7_Y___W.B__QEUK.ts` based on my version of `F_V7_Y___W.B__QEUK.js` from within the rethink container.  I've regenerated it by using the existing `F_V8_Y___W.B_2QEUK.ts` as a template.  Please see if this is suitable.

tests/cloud/devices/F_V7_Y___W.B__QEUK.test.ts
```diff
@@ -0,0 +1,134 @@
...
+import assert from 'node:assert/strict'
+import DUT from '@/cloud/devices/F_V7_Y___W.B__QEUK'
+import type { Metadata } from '@/cloud/thinq'
+import { MockHAConnection, MockThinq2Device, buf } from '@/tests/helpers/mocks'
+
+const DEVICE_ID = 'test-id'
+const MODEL_ID = 'F_V7_Y___W.B__QEUK'
+const META: Metadata = { modelId: MODEL_ID, modelName: 'F_V7_Y___W.B__QEUK', swVersion: '0.0.0' }
+
+// All hex dumps below are captured status telegrams from issue #75 (F6WV710P2S.ABLQPDG).
```
anszom, line 11: This test file appears to test your handler against captures made on a different device. This doesn't make much sense. The tests - unsurprisingly - fail

Testing against captures made on your own device would at least provide some evidence that the copy-pasted implementation is a good match.

anszom (changes requested):
The code is simple enough, but I will ask you to provide real tests. At some point we will need to do some cleanup of the copy-pasted implementations, but it is not strictly necessary now.

danleongjy:
Thank you for your guidance.  Yes I see that the the actual captures that I got from the rethink management interface are different from those reflected in the test for the other device.  I have rewritten the tests to take into account my actual captures.  I've also removed the Steam function that my washer does not have so that it need not be tested.

anszom:
squash-merged. Thanks :)

anszom:
Oops, I was too quick to pull the trigger. It appears that I've already added a handler for this device in #168.

@kroleg can you comment on this? You've assigned the `F_V8_Y___W_B_2QEUK` alias directly. Do the changes from this PR make sense for your device?

kroleg:
> TurboWash and Energy sensors as these are not featured on F_V7_Y___W.B__QEUK
 
I don't see turbowash on my device, but Energy is definitely present on my model

Screenshot 2026-09-18 at 19 48 00



Question is what's point of code duplication if alias works and you can just hide sensors which you don't need in Home Assistant

danleongjy:
Thanks for the feedback.  I implemented this based on my model: FV1409S4W, which reports as F_V7_Y___W.B__QEUK in HA but does not have energy reporting.  I did observe that the specification for F_V8_Y___W_B_2QEUK does work for F_V7_Y___W.B__QEUK, but thought that it is better to specify a dedicated handler that did not generate the unused sensors to avoid end-user confusion about why rethink reports sensor values for features that the physical model does not have.  But if we prefer the benefits of maintaining fewer handlers that cover more devices each, I'm happy to go with that approach too.

kroleg:
@danleongjy problem is that my F2V5PS0W also reports as `F_V7_Y___W.B__QEUK` but it has energy info (see https://github.com/anszom/rethink/pull/168). No idea why LG does that "same id - different sensors" thingy. Can probably add a note to readme explaining that this can happen and advice is to hide sensor which don't work / don't needed

danleongjy:
Sure that probably is the most practical way to deal with the many-to-one relationship between models and handlers. It does raise a question though of whether F_V8_Y___W_B_2QEUK is the most general case of all the models with similar capabilities. But probably we need more working configurations to probe the range of capabilities in existence to put us in a better position to refactor the handlers.

But since F_V7_Y___W.B__QEUK is a clean subset of F_V8_Y___W_B_2QEUK, I'm happy to go with the approach @kroleg has implemented.

PR #128, opened 2026-08-07
Map FV1413H2BA washer model F_VA_F___W.B__QEUK

BluSyn:
Register F_VA_F___W.B__QEUK (deviceType 201, RTK_RTL8711am) to the existing F_V__F___W.B_1QEUK front-load AABB washer handler.

Tested on LG model FV1413H2BA, and works perfectly with existing device support.


anszom:
Merged with a slight modification - I've added the device ID in README.md. Thanks!

PR #129, opened 2026-08-07
Add CST_570004_WW cassette AC support

BluSyn:
Map cassette model CST_570004_WW to the RAC TLV handler and accept 0xA7 fromDevice framing in tlv_device (required for this platform). Explicitly power on with non-off climate mode writes; unlock autodry, air purify, energy save, timers, and swing from state tags when caps bits are missing; expose 0x336 humidity sensor. Decode 0xA7 in packet-codec; autodry is a writable switch.

LG AC Cassette models tested: `ARNU36GTNA4` and `ARNU18GTQA4`. Internal model `CST_570004_WW` reported. These models don't have built-in wifi; was connected using wifi module accessory `PWFMDD200` through wifi module port on main board.

This has one unique feature: humidity reporting. All other features are generic to other devices as well.

This did require some refactoring due to difference in behavior vs base device. Tested base device on other compatible units without issue; no regressions found.

### DIFF
--- a/cloud/devices/RAC_056905_WW.ts
+++ b/cloud/devices/RAC_056905_WW.ts
@@ -264,6 +264,20 @@ export default class Device extends TLVDevice {
         return undefined
     }
 
+    /** True if this tag was present in a values/notify packet (value may be 0). */
+    hasTag(id: number) {
+        return this.raw_clip_state[id] != null
+    }
+
+    /**
+     * Optional features are unlocked by caps feature-words when present.
+     * Some models (e.g. CST_570004_WW cassette) omit or use different caps bits
+     * but still report the state tags — unlock from either signal.
+     */
+    hasCapOrTag(capWord: number, capBit: number, stateTag: number) {
+        return !!(this.raw_clip_state[capWord] & capBit) || this.hasTag(stateTag)
+    }
+
     initMakeSetConfig() {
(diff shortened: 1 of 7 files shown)


anszom:

cloud/devices/RAC_056905_WW.ts
```diff
@@ -264,6 +264,20 @@ export default class Device extends TLVDevice {
...
     }
 
+    /** True if this tag was present in a values/notify packet (value may be 0). */
+    hasTag(id: number) {
+        return this.raw_clip_state[id] != null
+    }
+
+    /**
+     * Optional features are unlocked by caps feature-words when present.
+     * Some models (e.g. CST_570004_WW cassette) omit or use different caps bits
```
maciejsszmigiero, line 274: I don't think we should enable features supported by the official ThinQ API and app just based on presence of particular tag - for example many (all?) `RAC_056905_WW` models have air purifier tag but only some models actually have an air purifier module inside.

Usually these features can be auto-detected from caps - that's how app knows whether to display particular feature widget and so should we.

Most likely `CST_570004_WW` simply has different cap bits than `RAC_056905_WW` - what does its model JSON file say?

anszom:
FYI, I'm deferring all AC work until a common base gets merged either in the form proposed in #122 or another. This may take some time though, I'm a bit overloaded right now.

PR #131, opened 2026-08-07
rethink-setup provisioning fixes for RTK_RTL8711am module

BluSyn:
May be hard to merge this since it touches important setup paths, but hoping it's useful for someone else.

This resolved 2 specific issues:
- `rethink-setup.ts` would succeed to provision device, but device doesn't fully join wifi after completion (this took lots of debugging to resolve, this PR adds lots of extra debugging output during setup as a result).
- After joining wifi, fails to connect to rethink due to weak TLS cipher support (legacy hardware or firmware)

Devices this resolved in testing:
- LG FV1413H2BA (front load washer)
- LG RVSX238MC (two-door fridge)

Both devices have `RTK_RTL8711am` modules, so likely hardware related. Combined these 2 fixes into one PR since they are likely related to same hardware support.

Specific changes:

- Improves rethink-setup setApInfo with a two-pass scan refine (frequency / security), multiprofile and legacy bssid fields, credential length logging, and delayed releaseDev. (delayed releaseDev is most likely the change that actually fixed the problem)

- Allow CBC-SHA cipher suites (SECLEVEL=0, min TLS 1.0) on device HTTPS/MQTTS listeners so older CLIP clients can complete /route request. (Note: nginx reverse proxies need the same ciphers if used to terminate 443)

### DIFF
--- a/rethink-cloud.ts
+++ b/rethink-cloud.ts
@@ -15,6 +15,7 @@ import { DeviceAcceptor as T2Acceptor } from './cloud/thinq2/device'
 import { Connection as HA_connection } from './cloud/homeassistant'
 import HA_bridge from './cloud/ha_bridge'
 import { normalize as normalizeConfig, RawConfig, CA } from './util/config'
+import { deviceTlsOptions } from './util/device_tls'
 import * as Management from './management'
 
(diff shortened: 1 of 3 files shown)


anszom:

PR #132, opened 2026-08-11
Add wt7305cv Washer

AndrewPaglusch:
Adds support for the LG WT7305CV washer

### DIFF
--- a/cloud/devices/T17A1EFHU_F.ts
+++ b/cloud/devices/T17A1EFHU_F.ts
@@ -1,0 +1,73 @@
+import HADevice from './base'
+import { Device as Thinq2Device } from '../thinq2/device'
+import { type Connection } from '../homeassistant'
+import { type Metadata } from '../thinq'
+import { allowExtendedType } from '@/util/casting'
+import AABBDevice from './aabb_device'
+
+const STATUS: Record<number, string> = {
+    0x00: 'Off',
+    0x02: 'Paused',
+    0x03: 'Sensing',
+    0x05: 'Wash',
+    0x06: 'Rinse',
+    0x07: 'Spin',
+    0x08: 'End',
+}
+
+export default class Device extends AABBDevice {
+    constructor(HA: Connection, thinq: Thinq2Device, meta: Metadata) {
+        super(HA, thinq)
+        this.setConfig(
+            allowExtendedType({
+                ...HADevice.config(meta, { name: 'LG Washer' }),
+                components: {
+                    power: {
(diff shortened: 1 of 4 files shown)


anszom:

cloud/devices/T17A1EFHU_F.ts
```diff
@@ -0,0 +1,73 @@
...
+        this.publishProperty('power', running ? 'ON' : 'OFF')
+        this.publishProperty('status', STATUS[phase] ?? 'unknown')
+        this.publishProperty('remaining_time', running ? mins : 0)
+    }
+
+    processAABB(buf: Buffer) {
+        if (buf[0] !== 0x20) return
+
+        // 0xDE: single status record (this model's equivalent of T1789's 0xEB/0xEC)
+        if (buf[1] === 0xde && buf.length >= 29) {
```
anszom, line 69: Does the packet length vary? If not, then it would be better to use a fixed length condition.

cloud/thinq2/device.ts
```diff
@@ -93,6 +93,10 @@ export class DeviceAcceptor extends TypedEmitter<DeviceAcceptorEvents> {
             }
 
             if (payload.cmd === 'device_packet' && payload.did === client.deployMsg?.did) {
+                // Some CLIP firmware never sends completeProvisioning_ack; register on the first packet
+                if (!client.deviceObj) {
+                    this.completeProvisioning(payload.did, payload, client)
+                }
```
anszom, line 99: Good find - although this approach could trigger some spurious warnings for devices that produce a completeProvisioning_ack packet late.
I've adjusted the code a bit and pushed on the T17A1EFHU_F-deploy-fix branch (303a95c). Please check if it works for you and rebase onto that branch.

cloud/devices/T17A1EFHU_F.ts
```diff
@@ -0,0 +1,73 @@
...
+                        unit_of_measurement: 'min',
+                    },
+                },
+            }),
+        )
+    }
+
+    private processRecord(rec: Buffer) {
+        const phase = rec[2]
+        const mins = rec[4]
```
anszom, line 57: maybe there is a `hours` field nearby?

anszom (commented):
Thanks for the PR. This implementation looks very similar to T1789EFH_F. Can you confirm that the discrepancies represent actual differences between the devices?

Also, please add a test file with captured packets to validate the implementation, see other devices for reference.

PR #136, opened 2026-08-13
Add support for F_C__Y___W.A__QEUK washing machine

martijndhondt:
## Summary
- Adds device support for the LG F_C__Y___W.A__QEUK front-loading washer (A-generation UK), decoding the 62-byte `0x20 0xEC` status packet, the compact 32-byte `0x20 0xEB` variant, and the 3-byte `0x20 0xD8` door-state packet.
- Exposes power, status, course, temp, spin, remote_start, door_lock, child_lock, steam, wrinkle_care, active, pre_state, tub_clean, initial/remaining/delay time, and error/error_message. All byte offsets were confirmed against live packet captures cross-checked against known machine settings and the Home Assistant UI, including a real `dE2` (door lock error) capture.
- Fixes `send_packet` in `cloud/thinq2/device.ts` to emit uppercase hex, which this device's firmware requires.

This was developed with the assistance of Claude (Anthropic) — packet captures were analysed and offsets cross-verified against known settings and HA UI state interactively.

## Test plan
- [x] `tests/cloud/devices/F_C__Y___W.A__QEUK.test.ts` — 38/38 passing, covering all decoded fields, the 0xD8 door-lock precedence rules, 0xE2 packet ignoring, and the error/error_message decoding
- [x] Full test suite run on this branch (excluding two pre-existing, unrelated Windows-only flaky tests already present on `master`)


anszom:
> excluding two pre-existing, unrelated Windows-only flaky tests already present on master

I'm not aware of those, but I haven't tried running on Windows. What are you referring to?

cloud/thinq2/device.ts
```diff
@@ -38,7 +38,7 @@ export class Device extends TypedEmitter<DeviceEvents> {
 
     send_packet(buf: Buffer) {
         this.emit('sendData', buf)
-        this.send('packet', 1, buf.toString('hex'))
+        this.send('packet', 1, buf.toString('hex').toUpperCase())
```
anszom, line 41: This sounds reasonable, but it should go in a separate commit.

cloud/devices/F_C__Y___W.A__QEUK.ts
```diff
@@ -0,0 +1,318 @@
...
+        if (prop === 'power') {
+            if (mqttValue === 'ON') {
+                this.send(Buffer.from('F02A0100', 'hex'))
+            } else if (mqttValue === 'OFF') {
+                this.send(Buffer.from('F024010100', 'hex'))
+            }
+        }
+
+        if (prop === 'pause') this.send(Buffer.from('F024040100', 'hex'))
+        if (prop === 'start') this.send(Buffer.from(mqttValue || 'F024050100', 'hex'))
```
anszom: How would an `mqttValue` appear here?

cloud/devices/F_C__Y___W.A__QEUK.ts
```diff
@@ -0,0 +1,318 @@
+import HADevice from './base'
+import { Device as Thinq2Device } from '../thinq2/device'
+import { type Connection } from '../homeassistant'
+import { type Metadata } from '../thinq'
+import { allowExtendedType } from '@/util/casting'
+import AABBDevice from './aabb_device'
+import { ERRORS, STATES, COURSES, TEMPERATURES, SPINS } from './washer_common'
+
+// F_C__Y___W.A__QEUK — LG front-loading washer, A-generation UK
```
anszom: Please add the official model name to the README as well.

martijndhondt:
Sorry, should've been more specific. Both are on master/upstream too (I checked with a clean upstream/master checkout before making any changes), just Windows-specific:

tests/bridge/util.test.ts → "subprocess rejects signals and spawn failures"
The test spawns a child Node process that calls process.kill(process.pid, 'SIGTERM') on itself, expecting bridge/util.ts (lines 66-78) to report terminated by SIGTERM. Windows has no real POSIX signal delivery — Node's docs note process.kill() on Windows unconditionally terminates the process rather than delivering an actual signal, so child_process's 'close' event reports signal: null, code: 1 instead of signal: 'SIGTERM'. That routes into the code !== 0 branch, producing "exited with code 1" instead of "terminated by SIGTERM". Documented Node/Windows limitation, not a bug in the util code.

tests/management/device-monitor.test.ts → "device monitor detaches device and manager listeners after a real WebSocket close"
Fails at line 39 — assert.equal(device.listenerCount('data'), 0) gets 1 instead. The test does ws.close(); await once(ws, 'close') then immediately asserts the server has already detached the device's listeners. On Windows the server-side close handling appears to lag behind the client seeing its own 'close' event — a client/server socket-close ordering race that Windows' networking stack surfaces more readily than Linux's. Since the assertion throws, server.close() on line 45 never runs, leaving the server listening — which is also why the suite takes noticeably longer to settle afterward.

Neither touches the washer device code, and both reproduce identically on a bare upstream/master checkout, so they're pre-existing and out of scope for this PR.

cloud/devices/F_C__Y___W.A__QEUK.ts
```diff
@@ -0,0 +1,318 @@
...
+        )
+    }
+
+    start() {
+        this.send(Buffer.from('F0ED1121010000001800', 'hex'))
+    }
+
+    processAABB(buf: Buffer) {
+        // Log every packet — remove once error/energy offsets are confirmed
+        console.log(`[F_C__Y___W.A__QEUK] AABB packet (${buf.length} bytes): ${buf.toString('hex')}`)
```
anszom: remove this please.

cloud/devices/F_C__Y___W.A__QEUK.ts
```diff
@@ -0,0 +1,318 @@
...
+            //                            live dE2 capture 2026-08-02: buf[10]=0x01 in all Error packets,
+            //                            buf[10]=0x00 in all non-error states (Off/Ready/Delayed/Measuring).
+            //   [22]    unknown          — varies; 0x03 in Off/Washing, 0x06 in Delayed/Spinning/End
+            //   [23]    pre_state        — last run state; mirrors buf[4] during active cycle;
+            //                            retains last state after power-off (e.g. End→Off transition shows End)
+            //   [25]    tub_clean        — 9 during wash; increments to 10 on End packet confirmed (NOTE: NOT buf[26])
+            // The 62-byte 0xEC packet has a second section [32..61] that during option
+            // selection holds the alternative configuration (the two sections alternate
+            // between current and previous settings as the user scrolls). Only [0..31]
+            // (first section) is read here; the second section is ignored.
```
anszom: simply put, the second section is the 'previous' state. Roughly half of the devices use 
`0xEC [new] [old]`
the others:
`0xEC [old] [new]`

cloud/devices/F_C__Y___W.A__QEUK.ts
```diff
@@ -0,0 +1,318 @@
...
+            const remain_h = buf[5] // remaining_time hours (counts down; 0 briefly during load-measuring)
+            const remain_m = buf[6] // remaining_time minutes
+            const initial_h = buf[7] // initial_time hours (fixed for the lifetime of the program)
+            const initial_m = buf[8] // initial_time minutes
+            const lock_status = buf[9]
+            const error_code = buf[10]
+            const spin = buf[12]
+            const temp = buf[13] // 0x00 during rinse (cold water) → publishes 'unknown', expected
+            const course = buf[14]
+            const delay_h = buf[16]
```
anszom: this diverges from other washers, are you sure that the fields are decoded correctly? I admit that I have no clear evidence that something is wrong here, but I'm using Claude as a review aid and it is very nervous about this code :) See also my comments about tests

tests/cloud/devices/F_C__Y___W.A__QEUK.test.ts
```diff
@@ -0,0 +1,482 @@
...
+// Frame format: AA <total_len> 20 EC <payload...> <checksum> BB (62-byte 0xEC inner block = 66 bytes total).
+// Confirmed offsets — see device file header.
+
+// Real capture: Cotton/60°C/1400 RPM, 104 min remaining, 121 min initial, delay=0, tub_clean=10.
+// buf[18]=0x00 → steam=OFF.
+const SAMPLE_WASHING_EC = buf(
+    'AA4220EC001C06012C02010100030A0601000000004000000306000A003400000500001C06012B02010100030A0601000000004000000306000A003400000500FEBB',
+)
+
+// Derived from SAMPLE_WASHING_EC with raw[20] (= buf[18]) changed 0x00→0x80 to set the steam flag.
```
anszom: Real captured samples would be much more useful, otherwise we are validating against our own assumptions.

anszom (changes requested):
Thank  you for the PR and for your patience :) See my comments above. Also, please add the public name of the device to README.

cloud/thinq2/device.ts
martijndhondt: done, it's its own commit now (thinq2: send packet hex in uppercase for device compatibility).

cloud/devices/F_C__Y___W.A__QEUK.ts
martijndhondt: added: F4WV709P1 (F_C__Y___W.A__QEUK).

cloud/devices/F_C__Y___W.A__QEUK.ts
martijndhondt: Removed the debug console.log

cloud/devices/F_C__Y___W.A__QEUK.ts
martijndhondt: you were exactly right. I compared consecutive real captures byte-for-byte and confirmed the 62-byte 0xEC packet is [old][new], not [current][alternate] like I'd assumed. Proof: record 2 of packet N always equals record 1 of packet N+1, across three independent sequences — normal 1-minute polling, a rapid option-scroll, and a real dE2 error transition (record 2 showed Error three seconds before record 1 caught up). I'd been reading record 1 (the stale one) this whole time, so every published value was one packet behind — not obviously wrong in the UI since it's usually just a few seconds to ~1 minute of lag, but wrong. Fixed by extracting the correct current record (buf[32:62] for 0xEC, buf[2:32] for 0xEB, which has no old/new pair) into a shared processRecord(), matching the pattern F3L2CYU__ and T1789EFH_F already use in this codebase.

tests/cloud/devices/F_C__Y___W.A__QEUK.test.ts
martijndhondt: recomputed every "real capture" fixture against the corrected (record 2) reading rather than what record 1 showed, so several expected values changed even though the captured bytes didn't (e.g. the remote_start fixture turned out to demonstrate a different course/spin than I'd originally asserted, once read from the right record). I also went back through a ~34k-line capture log and found a real steam-ON packet to replace the synthetic one. I could not find a real wrinkle_care-ON or child-lock-engaged capture in that log — the machine apparently never had those settings on during any session it was captured in. Those two fixtures remain synthetic bit-flips, now clearly labeled as such with a note to replace them once I can toggle those settings live (can't right now — will follow up).

cloud/devices/F_C__Y___W.A__QEUK.ts
martijndhondt: this was the section-order bug. With the fix, the record-relative field layout is the same shape as those other devices use.

martijndhondt:
Small follow-up: got a real child-lock-engaged capture (2026-09-04) and replaced the synthetic bit-flip fixture with it. wrinkle_care-ON is still synthetic — the machine hasn't run with that setting on yet.

cloud/devices/F_C__Y___W.A__QEUK.ts
```diff
@@ -0,0 +1,321 @@
...
+                        options: STATES.filter((a) => a !== undefined),
+                    },
+                    tub_clean: {
+                        platform: 'sensor',
+                        unique_id: '$deviceid-tub-clean',
+                        state_topic: '$this/tub_clean',
+                        name: 'Tub clean counter',
+                        icon: 'mdi:washing-machine-alert',
+                        entity_category: 'diagnostic',
+                    },
```
anszom: For consistency with other implementations, please rename this to tub_clean_count - [see here](https://github.com/anszom/rethink/blob/b41950d91a717fd85de7794c1de25b9c1bdae1ba/cloud/devices/F3L7CYK5W_US_WIFI.ts#L357-L365).

cloud/devices/F_C__Y___W.A__QEUK.ts
```diff
@@ -0,0 +1,321 @@
...
+                        state_topic: '$this/door_lock',
+                        name: 'Door lock',
+                        device_class: 'lock',
+                    },
+                    steam: {
+                        platform: 'binary_sensor',
+                        unique_id: '$deviceid-steam',
+                        state_topic: '$this/steam',
+                        name: 'Steam',
+                        icon: 'mdi:weather-fog',
```
anszom: for consistency use: `mdi:kettle-steam`

cloud/devices/F_C__Y___W.A__QEUK.ts
```diff
@@ -0,0 +1,321 @@
...
+                    },
+                    wrinkle_care: {
+                        platform: 'binary_sensor',
+                        unique_id: '$deviceid-wrinkle_care',
+                        state_topic: '$this/wrinkle_care',
+                        name: 'Wrinkle care',
+                        icon: 'mdi:iron-outline',
```
anszom: consistency: use `mdi:tshirt-crew-outline`.

cloud/devices/F_C__Y___W.A__QEUK.ts
```diff
@@ -0,0 +1,321 @@
...
+                    },
+                    pre_state: {
+                        platform: 'sensor',
+                        unique_id: '$deviceid-pre_state',
+                        state_topic: '$this/pre_state',
+                        name: 'Pre state',
+                        icon: 'mdi:state-machine',
+                        device_class: 'enum',
+                        options: STATES.filter((a) => a !== undefined),
+                    },
```
anszom: Is there a good reason to expose this to HA? In all other devices we simply ignore the previous state.

cloud/devices/F_C__Y___W.A__QEUK.ts
```diff
@@ -0,0 +1,321 @@
...
+        this.publishProperty('steam', steam ? 'ON' : 'OFF')
+        this.publishProperty('wrinkle_care', wrinkle_care ? 'ON' : 'OFF')
+        this.publishProperty('active', active ? 'ON' : 'OFF')
+        this.publishProperty('child_lock', child_lock ? 'OFF' : 'ON')
+        this.publishProperty('pre_state', STATES[pre_state] ?? 'unknown')
+        this.publishProperty('tub_clean', tub_clean)
+
+        // Derive door_lock from status for Delayed and active-cycle states where
+        // 0xD8 packets are not emitted. HA device_class='lock': OFF=Locked, ON=Unlocked.
+        // Off(0) → unlocked → ON; Ready(1) → 0xD8 authoritative; everything else → locked → OFF.
```
anszom, line 254: Are you sure that the status packet doesn't have a bit for the door lock state? I'd double-check this.

If it's actually missing then this is the best we can do...

anszom (changes requested):
Just some minor fixes&questions, see below

martijndhondt:
Thanks — addressed all five:

1. Renamed tub_clean → tub_clean_count (matches VCDWL2QEUK / F3L7CYK5W_US_WIFI).
2. Icons: steam → mdi:kettle-steam, wrinkle_care → mdi:tshirt-crew-outline.
3. Removed the pre_state entity entirely. You're right — no other device here exposes the previous-state half of a dual-record packet (STUDIO_HOOD documents the same [previous][current] layout and only reads current). The byte stays as an internal, unmapped comment.
4. Door-lock bit: took this seriously and actually checked rather than assumed. Scanned the full capture log for every 0xD8 lock/unlock transition (91 total), then isolated the 10 that happened while status stayed Ready with no simultaneous course/spin/temp scroll (to rule out contamination), and diffed the EC record byte-for-byte across each transition. No byte changes consistently with lock state. The byte that changes most often (offset 20, previously just "unknown") moves unpredictably — 1→0 in some transitions, 2→3 or 4→4 in others — which looks like dial-scroll noise, not a lock flag. Two other candidate bytes never changed at all across any transition. I'm fairly confident this model genuinely has no door-lock bit in the status record and 0xD8 is the only real-time source. Documented this inline with the evidence so it's not just an assumption anymore.

anszom:
Looks good, squash-merged. Thanks for the contribution!

PR #139, opened 2026-08-15
feat(dishwasher): Add LG Dishwasher (H11) support

asata:
This PR adds support for the LG Dishwasher to the local AABB packet parser.
- **Actual Model Name**: DUE2BG.AKOR
- **LG ThinQ / Rethink Internal Model ID**: H11

It includes read-only sensors for monitoring the dishwasher's state and full remote control capabilities reverse-engineered from the ThinQ app, including comprehensive auxiliary settings.

### Features Added
- **State Monitoring**: `state`, `course`, `remain_time`
- **Basic Control**: Power `switch`, Pause `button`, Resume `button`, Cancel / Drain Stop `button`
- **Remote Start Options**: 
  - Allows selecting Target Course (Auto, 1 Hour, Normal, Download Cycle, etc.)
  - Delay Start Hour (1~12 hours)
  - Washing Options: High Temp (`0x08`), Extra Dry (`0x04`), Extra Rinse (1~3 levels)
  - `start_course` button dynamically constructs the `f0 26 10` control payload based on the selected virtual entities.
- **Comprehensive Settings Control**: 
  - Added 7 entities for configuring auxiliary device settings directly from HA: `salt_level`, `rinse_level`, `buzzer_level`, `end_alarm_sound`, `clean_reminder`, `auto_dry`, and `brightness`.
  - Added `remote_start_mode` (Permanent, One-Time, Off) to configure the machine's behavior after cycles.

---

### 🔍 Complete Packet Analysis & Protocol Structure (For Reviewers)

During the development of this parser, we successfully decoded the complete structure of the AABB local protocol for both status reporting and appliance control. Here is the full breakdown of the protocol for the H11 Dishwasher:

#### 1. Status Payload (`32 ec` -> `00 18` Tag Block)
The machine reports its current state using `32 ec` packets. The second half of the payload contains a `00 18` tag block followed by 24 bytes of data. Key offsets (starting at 0):
- **`[0]` (State)**: `01` (INITIAL), `02` (RUNNING), `03` (PAUSE), `04` (STANDBY)
- **`[3] & [4]` (Initial Time)**: Initial Course Time (Hour, Minute)
- **`[5]` (Course)**: Base Course Code (e.g., `01`: Auto, `12`: 1 Hour, `0b`: Download Cycle)
- **`[7] & [8]` (Remain Time)**: Remaining Time (Hour, Minute)
- **`[9]` (Delay Start)**: Delay start time in hours (`0` means disabled)
- **`[11]` (Opt1 & Door)**: `0x40` (Clean Reminder LED ON), `0x10` (Auto Dry ON), `0x02` (Door Open)
- **`[12]` (Wash Options)**: `0x04` (Extra Dry ON), `0x08` (High Temp ON)
- **`[13]` (Rinse Level)**: Rinse dispenser level (`0x00` ~ `0x04`)
- **`[14]` (Salt Level)**: Salt dispenser level (`0x00` ~ `0x04`)
- **`[15]` (Buzzer & Btn)**: `0x80` (Buzzer HIGH), `0x40` (Buzzer LOW), `0x00` (Buzzer OFF). *(Note: `0x02` represents the physical Remote Start button state)*
- **`[16]` (Opt2)**: `0x80` (Remote Control: PERMANENT), `0x40` (ONE-TIME), `0xc0` (OFF). `0x04` (End Alarm Sound ON)
- **`[19]` (Opt3)**: `0x40` (Display Brightness HIGH)
- **`[20]` (Smart Course)**: Specific Downloaded Course ID (e.g., `0x0f`: Plastic, `0x0d`: Tub Clean)
- **`[21]` (Extra Rinse)**: `00` (0), `10` (1 level), `20` (2 levels), `30` (3 levels)

#### 2. Control Commands (`f0 26`)
Unlike washing machines (`32 0a`), the dishwasher uses `f0 26` packets for control.
- **Power On / Wake Up**: `f0 26 16`
- **Power Off (Immediate)**: `f0 26 12`
- **Pause / Resume**: `f0 26 13` / `f0 26 14`
- **Cancel Course & Drain**: `f0 26 11` *(Pumps for 1 min, then powers off)*
- **Start Course**: `f0 26 10 [Course] [DelayHour] [Opt2=0x00] [Opt3] [Opt4] [Opt5=0x00]`
  - `[Opt3]` (Wash Options): `0x08` (High Temp), `0x04` (Extra Dry). Both = `0x0c`.
  - `[Opt4]` (Rinse Levels): `0x08` (1), `0x10` (2), `0x18` (3). *(Note: The `0x40` flag must be added if the course is `0b` Download Cycle)*
- **Settings Control**: `f0 26 [Rinse] [Salt] [Opt1] [Opt2] [Opt3] 00 00 00 00`
  - Overwrites all auxiliary settings at once. Our implementation caches existing user preferences and applies bitwise operations to update only the requested setting without altering others.

#### 3. Note on "Downloaded Courses"
Changing the specific "Downloaded Course" assignment (e.g., changing from *Tub Clean* to *Pots & Pans*) **cannot** be executed via the local network. When changed via the app, the machine downloads the cycle algorithm payload directly from LG Cloud (HTTPS/MQTT). We can monitor the active course ID by reading byte `[20]`, but we cannot push a new assignment locally.

### DIFF
--- a/cloud/devices/H11.ts
+++ b/cloud/devices/H11.ts
@@ -1,0 +1,232 @@
+import HADevice from './base'
+import { Device as Thinq2Device } from '../thinq2/device'
+import { DeviceDiscovery, type Connection } from '../homeassistant'
+import { type Metadata } from '../thinq'
+import { allowExtendedType } from '@/util/casting'
+import AABBDevice from './aabb_device'
+
+const DISHWASHER_STATES: Record<number, string> = {
+    1: 'INITIAL',
+    2: 'RUNNING',
+    3: 'PAUSE',
+    4: 'STANDBY',
+}
+
+const COURSES: Record<number, string> = {
+    0x00: 'OFF',
+    0x01: 'AUTO',
+    0x12: 'ONE_HOUR',
+    0x05: 'NORMAL/ECO',
+    0x02: 'HEAVY/INTENSIVE',
+    0x10: 'SILENT_NIGHT',
+    0x08: 'EXPRESS',
+    0x0b: 'DOWNLOAD_CYCLE',
+    0x09: 'MACHINE_CLEAN',
+}
+
+const SMART_COURSES: Record<number, string> = {
+    0x05: 'GREASY_TABLEWARE',
+    0x0d: 'MACHINE_CLEAN',
+    0x0f: 'PLASTIC_WASH',
+}
+
+const RINSE_LEVELS: Record<number, string> = {
+    0x00: 'OFF',
+    0x10: 'LEVEL_1',
+    0x20: 'LEVEL_2',
+    0x30: 'LEVEL_3',
+}
+
+export default class Device extends AABBDevice {
+    readonly deviceConfig: DeviceDiscovery
+
+    constructor(HA: Connection, thinq: Thinq2Device, meta: Metadata) {
+        super(HA, thinq)
+        this.deviceConfig = HADevice.config(meta, { name: 'LG Dishwasher' })
+
+        this.setConfig(
+            allowExtendedType({
+                ...this.deviceConfig,
+                components: {
+                    power: {
+                        platform: 'switch',
+                        unique_id: '$deviceid-power',
+                        state_topic: '$this/power',
+                        command_topic: '$this/power/set',
+                        name: 'Power',
+                        icon: 'mdi:power',
+                    },
+                    state: {
+                        platform: 'sensor',
+                        icon: 'mdi:washing-machine',
+                        unique_id: '$deviceid-state',
+                        state_topic: '$this/state',
+                        name: 'State',
+                    },
+                    course: {
+                        platform: 'sensor',
+                        icon: 'mdi:dishwasher',
+                        unique_id: '$deviceid-course',
+                        state_topic: '$this/course',
+                        name: 'Course',
+                    },
+                    remain_time: {
+                        platform: 'sensor',
+                        icon: 'mdi:timer-sand',
+                        unique_id: '$deviceid-remain_time',
+                        state_topic: '$this/remain_time',
+                        name: 'Remain Time',
+                        unit_of_measurement: 'min',
+                    },
+                    course_time: {
+                        platform: 'sensor',
+                        icon: 'mdi:timer',
+                        unique_id: '$deviceid-course_time',
+                        state_topic: '$this/course_time',
+                        name: 'Course Time',
+                        unit_of_measurement: 'min',
+                    },
+                    door: {
+                        platform: 'binary_sensor',
+                        device_class: 'door',
+                        unique_id: '$deviceid-door',
+                        state_topic: '$this/door',
+                        name: 'Door',
+                        payload_on: 'OPEN',
+                        payload_off: 'CLOSE',
+                    },
+                    high_temp_dry: {
+                        platform: 'binary_sensor',
+                        icon: 'mdi:weather-sunny',
+                        unique_id: '$deviceid-high_temp_dry',
+                        state_topic: '$this/high_temp_dry',
+                        name: 'High Temp Dry',
+                        payload_on: 'ON',
+                        payload_off: 'OFF',
+                    },
+                    sterilize: {
+                        platform: 'binary_sensor',
+                        icon: 'mdi:thermometer-high',
+                        unique_id: '$deviceid-sterilize',
+                        state_topic: '$this/sterilize',
+                        name: 'Sterilize',
+                        payload_on: 'ON',
+                        payload_off: 'OFF',
+                    },
+                    rinse_level: {
+                        platform: 'sensor',
+                        icon: 'mdi:water-plus',
+                        unique_id: '$deviceid-rinse_level',
+                        state_topic: '$this/rinse_level',
+                        name: 'Rinse Level',
+                    },
+                    delay_start: {
+                        platform: 'sensor',
+                        icon: 'mdi:clock-fast',
+                        unique_id: '$deviceid-delay_start',
+                        state_topic: '$this/delay_start',
+                        name: 'Delay Start (Hours)',
+                        unit_of_measurement: 'h',
+                    },
+                    remote_start: {
+                        platform: 'binary_sensor',
+                        icon: 'mdi:remote',
+                        unique_id: '$deviceid-remote_start',
+                        state_topic: '$this/remote_start',
+                        name: 'Remote Start',
+                        payload_on: 'ON',
+                        payload_off: 'OFF',
+                    },
+                },
+            }),
+        )
+    }
+
+    setProperty(prop: string, mqttValue: string) {
+        if (prop === 'power') {
+            if (mqttValue === 'ON') {
+                this.send(Buffer.from('F02616', 'hex')) // Wake Up
+            } else if (mqttValue === 'OFF') {
+                this.send(Buffer.from('F02612', 'hex')) // Immediate Power Off
+            }
+        }
+    }
+
+    processAABB(buf: Buffer) {
+        if (buf[0] === 0x32 && buf[1] === 0xec) {
+            const payloadLen = buf.length - 2
+            const halfLen = Math.floor(payloadLen / 2)
+            if (halfLen > 10) {
+                const curStatus = buf.subarray(2 + halfLen, buf.length)
+                this.processStatus(curStatus)
+            }
+        }
+    }
+
+    processStatus(curStatus: Buffer) {
+        if (curStatus[0] === 0x00 && curStatus[1] === 0x18) {
+            const data = curStatus.subarray(2, 26) // 24 bytes
+
+            const stateCode = data[0]
+            const processCode = data[1]
+            const stateStr = DISHWASHER_STATES[stateCode] || `UNKNOWN(${stateCode})`
+
+            // 전원 상태는 STANDBY(4) 이거나 취소 중(processCode: 0x63)일 때 OFF로 처리하여 스위치 튕김 방지
+            const isPowerOff = stateCode === 4 || processCode === 0x63
+
+            this.publishProperty('state', stateStr)
+            this.publishProperty('power', isPowerOff ? 'OFF' : 'ON')
+
+            // Course (Index 5)
+            const baseCourseCode = data[5]
+            const smartCourseCode = data[20]
+
+            let courseStr = ''
+            if (smartCourseCode !== 0) {
+                // If a smart course is active, use it instead of the base course
+                courseStr =
+                    SMART_COURSES[smartCourseCode] ||
+                    `DOWNLOAD_COURSE(0x${smartCourseCode.toString(16).padStart(2, '0')})`
+            } else {
+                courseStr = COURSES[baseCourseCode] || `UNKNOWN(0x${baseCourseCode.toString(16).padStart(2, '0')})`
+            }
+
+            this.publishProperty('course', courseStr)
+
+            // Initial(Course) Time (Index 3: hour, Index 4: minute)
+            const initialHour = data[3]
+            const initialMinute = data[4]
+            this.publishProperty('course_time', initialHour * 60 + initialMinute)
+
+            // Remain Time (Index 7: hour, Index 8: minute)
+            const remainHour = data[7]
+            const remainMinute = data[8]
+            this.publishProperty('remain_time', remainHour * 60 + remainMinute)
+
+            // Delay Start Hour (Index 9)
+            const delayHour = data[9]
+            this.publishProperty('delay_start', delayHour)
+
+            // Door (Index 11 bit 0x02)
+            const isDoorOpen = (data[11] & 0x02) !== 0
+            this.publishProperty('door', isDoorOpen ? 'OPEN' : 'CLOSE')
+
+            // High Temp Dry (Index 12 bit 0x04)
+            const isExtraDry = (data[12] & 0x04) !== 0
+            this.publishProperty('high_temp_dry', isExtraDry ? 'ON' : 'OFF')
+
+            // Sterilize (Index 12 bit 0x08)
+            const isHighTemp = (data[12] & 0x08) !== 0
+            this.publishProperty('sterilize', isHighTemp ? 'ON' : 'OFF')
+
+            // Remote Start (Index 15 bit 0x02)
+            const isRemoteStart = (data[15] & 0x02) !== 0
+            this.publishProperty('remote_start', isRemoteStart ? 'ON' : 'OFF')
+
+            // Rinse Level (Index 21)
+            const rinseCode = data[21]
+            const rinseStr = RINSE_LEVELS[rinseCode] || `UNKNOWN(${rinseCode})`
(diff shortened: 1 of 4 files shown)


anszom:

cloud/devices/H11.ts
```diff
@@ -0,0 +1,579 @@
...
+                    },
+                    cancel: {
+                        platform: 'button',
+                        icon: 'mdi:stop',
+                        unique_id: '$deviceid-cancel',
+                        command_topic: '$this/cancel/set',
+                        name: 'Cancel / Drain Stop',
+                        payload_press: 'PRESS',
+                    },
+                    target_course: {
```
anszom, line 353: A full set of target/current entities is not very user-friendly. Can you explain why this is the best approach?

cloud/devices/H11.ts
```diff
@@ -0,0 +1,579 @@
...
+                        options: [
+                            'AUTO',
+                            'ONE_HOUR',
+                            'NORMAL/ECO',
+                            'HEAVY/INTENSIVE',
+                            'SILENT_NIGHT',
+                            'EXPRESS',
+                            'DOWNLOAD_CYCLE',
+                            'MACHINE_CLEAN',
+                        ],
```
anszom: This looks like a copy of the `COURSES` keys, so it should be derived from there instead of copy/pasted.

cloud/devices/H11.ts
```diff
@@ -0,0 +1,579 @@
...
+    0x0d: 'MACHINE_CLEAN',
+    0x0f: 'PLASTIC_WASH',
+}
+
+const RINSE_LEVELS: Record<number, string> = {
+    0x00: 'OFF',
+    0x10: 'LEVEL_1',
+    0x20: 'LEVEL_2',
+    0x30: 'LEVEL_3',
+}
```
anszom, line 38: These should be user-readable labels, so let's use natural formatting: 'Initial', 'One hour', 'Level 1' and so on.

cloud/devices/H11.ts
```diff
@@ -0,0 +1,579 @@
...
+
+    processStatus(curStatus: Buffer) {
+        if (curStatus[0] === 0x00 && curStatus[1] === 0x18) {
+            const data = curStatus.subarray(2, 26) // 24 bytes
+
+            const stateCode = data[0]
+            const processCode = data[1]
+            const stateStr = DISHWASHER_STATES[stateCode] || `UNKNOWN(${stateCode})`
+
+            // 전원 상태는 STANDBY(4) 이거나 취소 중(processCode: 0x63)일 때 OFF로 처리하여 스위치 튕김 방지
```
anszom: Please keep comments in English:)

cloud/devices/H11.ts
```diff
@@ -0,0 +1,579 @@
...
+            const coursesReverse: Record<string, number> = {
+                AUTO: 0x01,
+                ONE_HOUR: 0x12,
+                'NORMAL/ECO': 0x05,
+                'HEAVY/INTENSIVE': 0x02,
+                SILENT_NIGHT: 0x10,
+                EXPRESS: 0x08,
+                DOWNLOAD_CYCLE: 0x0b,
+                MACHINE_CLEAN: 0x09,
+            }
```
anszom: This should be derived from `COURSES` too.

cloud/devices/H11.ts
```diff
@@ -0,0 +1,579 @@
...
+            const isDoorOpen = (data[11] & 0x02) !== 0
+            this.publishProperty('door', isDoorOpen ? 'OPEN' : 'CLOSE')
+
+            // High Temp Dry (Index 12 bit 0x04)
+            const isExtraDry = (data[12] & 0x04) !== 0
+            this.publishProperty('high_temp_dry', isExtraDry ? 'ON' : 'OFF')
+
+            // Sterilize (Index 12 bit 0x08)
+            const isHighTemp = (data[12] & 0x08) !== 0
+            this.publishProperty('sterilize', isHighTemp ? 'ON' : 'OFF')
```
anszom: The comments don't match the code.

cloud/devices/H11.ts
```diff
@@ -0,0 +1,579 @@
...
+    }
+
+    processAABB(buf: Buffer) {
+        if (buf[0] === 0x32 && buf[1] === 0xec) {
+            const payloadLen = buf.length - 2
+            const halfLen = Math.floor(payloadLen / 2)
+            if (halfLen > 10) {
+                const curStatus = buf.subarray(2 + halfLen, buf.length)
+                this.processStatus(curStatus)
+            }
```
anszom: Instead of doing arithmetic on the packet length, can we just compare it to a predefined value? Do you expect packets of a varying length here? See other devices for an example of the xxEC / xxEB pattern.

cloud/devices/H11.ts
```diff
@@ -0,0 +1,579 @@
...
+                    power: {
+                        platform: 'switch',
+                        unique_id: '$deviceid-power',
+                        state_topic: '$this/power',
+                        command_topic: '$this/power/set',
+                        name: 'Power',
+                        icon: 'mdi:power',
+                    },
+                    state: {
+                        platform: 'sensor',
```
anszom, line 85: Use `class: 'enum'`

cloud/devices/H11.ts
```diff
@@ -0,0 +1,579 @@
...
+        this.publishProperty('energy_consumption', energyAccum.toString())
+    }
+
+    processStatus(curStatus: Buffer) {
+        if (curStatus[0] === 0x00 && curStatus[1] === 0x18) {
+            const data = curStatus.subarray(2, 26) // 24 bytes
+
+            const stateCode = data[0]
+            const processCode = data[1]
+            const stateStr = DISHWASHER_STATES[stateCode] || `UNKNOWN(${stateCode})`
```
anszom: After changing the entity to `enum`, the fallback will have to be simply 'unknown'

cloud/devices/H11.ts
```diff
@@ -0,0 +1,579 @@
...
+                    },
+                    state: {
+                        platform: 'sensor',
+                        icon: 'mdi:washing-machine',
+                        unique_id: '$deviceid-state',
+                        state_topic: '$this/state',
+                        name: 'State',
+                    },
+                    course: {
+                        platform: 'sensor',
```
anszom, line 97: I'm not sure how the downloaded course names work. If the list is bounded, we should use an `enum` here as well.

cloud/devices/H11.ts
```diff
@@ -0,0 +1,579 @@
...
+                        payload_off: 'OFF',
+                    },
+                    target_extra_rinse: {
+                        platform: 'select',
+                        icon: 'mdi:water-plus',
+                        unique_id: '$deviceid-target_extra_rinse',
+                        state_topic: '$this/target_extra_rinse',
+                        command_topic: '$this/target_extra_rinse/set',
+                        name: 'Extra Rinse',
+                        options: ['0', '1', '2', '3'],
```
anszom: A slight inconsistency: this is an enum, while `rinse_level` and `salt_level` use a ranged number entity.

anszom (changes requested):
Thanks for the contribution. Please see my comments below. The most controversial area is the target/current entity doubling.

Please add a test file with real packets captured for this appliance. Also, please add the dishwasher's model name (DUE2BG.AKOR) into the README.md list.

cloud/devices/H11.ts
asata: Home Assistant currently lacks a single unified service to start a dishwasher with specific options (like High Temp, Extra Rinse, etc.) all at once. Because of this, I had to expose target virtual entities so users can configure the options beforehand and assemble the complete payload when the 'Start' button is pressed.

If we were to combine the current state and target state into a single entity (e.g., a single select entity for course), changing the value in HA wouldn't actually change the physical appliance's state until the 'Start' command is sent. This would cause a critical mismatch between the HA UI and the actual machine state. Also, it would be problematic if the user tries to change the dropdown while the machine is already running. Therefore, separating the 'Target' (options to be sent on start) and 'Current' (what the machine is actually doing) is the safest approach to prevent UI synchronization issues.

cloud/devices/H11.ts
```diff
@@ -0,0 +1,715 @@
...
+                    // Current operating state of the dishwasher.
+                    // data[0]: 01=INITIAL, 02=RUNNING, 03=PAUSE, 04=STANDBY
+                    state: {
+                        platform: 'sensor',
+                        device_class: 'enum',
+                        icon: 'mdi:washing-machine',
+                        unique_id: '$deviceid-state',
+                        state_topic: '$this/state',
+                        name: 'State',
+                        options: Array.from(new Set([...Object.values(DISHWASHER_STATES), 'unknown'])),
```
anszom, line 91: 'unknown' is assumed by HA, no need to state it here. The same applies to all other `options:`

tests/cloud/devices/H11.test.ts
```diff
@@ -0,0 +1,122 @@
...
+import assert from 'node:assert/strict'
+import DUT from '@/cloud/devices/H11'
+import type { Metadata } from '@/cloud/thinq'
+import { MockHAConnection, MockThinq2Device, buf } from '@/tests/helpers/mocks'
+
+const DEVICE_ID = 'test-id'
+const MODEL_ID = 'H11'
+const META: Metadata = { modelId: MODEL_ID, modelName: 'H11', swVersion: '0.0.0' }
+
+// Mock 32ec packet (54 bytes)
```
anszom, line 11: Please test on actual captured data.

tests/cloud/devices/H11.test.ts
```diff
@@ -0,0 +1,122 @@
...
+// 49: Opt3 (0x40 = Brightness High)
+// 50: Smart Course (00 = none)
+// 51: Extra Rinse (10 = 1 level)
+// 52, 53: unused
+const RUNNING_STATUS = buf(
+    '00000000000000000000000000000000000000000000000000000000001802000001300100010f0000400c0203808000004000100000'
+)
+
+describe('H11 Dishwasher', () => {
+    test('parses RUNNING status correctly', () => {
```
anszom, line 40: This test fails

cloud/devices/H11.ts
```diff
@@ -0,0 +1,715 @@
...
+            this.send(Buffer.from([0xf0, 0x26, 0x10, this.targetCourse, this.targetDelay, 0x00, opt3, opt4, 0x00]))
+        }
+    }
+
+    processAABB(buf: Buffer) {
+        if (buf[0] === 0x32 && buf[1] === 0xec) {
+            // Scan for '00 18' status TLV blocks starting at offset 2.
+            // Older firmware sends 54-byte packets (two 26-byte blocks).
+            // Newer firmware sends larger packets (e.g. 94 bytes) with additional
+            // TLV data appended after each 24-byte status block.
```
anszom, line 566: This doesn't sound like a reliable approach. 
1. `00 18` could appear somewhere in the middle of a status block
2. Even if it doesn't there will likely be two blocks, each containing '00 18', so you will run `processStatus` twice

Also, since you aim to support two firmwares/formats here, you should provide test data for both - actual data captured from an appliance. If you don't have a device with the old firmware anymore, then just stick to supporting the new format.

For a single format, [this](https://github.com/anszom/rethink/blob/b41950d91a717fd85de7794c1de25b9c1bdae1ba/cloud/devices/2REF11EBIVPC4.ts#L102-L111) is the preferred approach. To support two protocol variants, you may want to have four `if`s instead of two.

BTW, this format doesn't resemble TLV (type-length-value) encoding, so the comment is misleading.

cloud/devices/H11.ts
```diff
@@ -0,0 +1,715 @@
...
+
+            // Delay Start Hour (Index 9)
+            const delayHour = data[9]
+            this.publishProperty('delay_start', delayHour)
+
+            // Door (Index 11 bit 0x02)
+            const isDoorOpen = (data[11] & 0x02) !== 0
+            this.publishProperty('door', isDoorOpen ? 'OPEN' : 'CLOSE')
+
+            // 추가 건조 (Index 12 bit 0x04): 세척 후 고열 추가 건조 활성 여부
```
anszom, line 659: Please keep comments in English.

tests/cloud/devices/H11.test.ts
```diff
@@ -0,0 +1,122 @@
...
+
+        // Configure cached settings first
+        dut.cachedRinseLevel = 0x02
+        dut.cachedSaltLevel = 0x03
+        dut.cachedAutoDry = true
+        dut.cachedCleanReminder = false
+        dut.cachedBuzzerLevel = 'HIGH'
+        dut.cachedRemoteStartMode = 'PERMANENT'
+        dut.cachedEndAlarmSound = false
+        dut.cachedBrightness = true
```
anszom, line 88: this fails typecheck

cloud/devices/H11.ts
```diff
@@ -0,0 +1,715 @@
...
+                    },
+                    // Current operating state of the dishwasher.
+                    // data[0]: 01=INITIAL, 02=RUNNING, 03=PAUSE, 04=STANDBY
+                    state: {
+                        platform: 'sensor',
+                        device_class: 'enum',
+                        icon: 'mdi:washing-machine',
+                        unique_id: '$deviceid-state',
+                        state_topic: '$this/state',
+                        name: 'State',
```
anszom, line 90: Please adapt the naming, classes & icons to better match other devices. For example:
'Status' instead of 'State, 'Remaining time' instead of 'Remain Time' and so on.

cloud/devices/H11.ts
anszom: ok, sounds reasonable.

cloud/devices/H11.ts
```diff
@@ -0,0 +1,715 @@
...
+                    // Opt2 bits 0x80=PERMANENT, 0x40=ONE_TIME, 0xc0=OFF.
+                    // Read back from data[16] bits 0xc0.
+                    remote_start_mode: {
+                        platform: 'select',
+                        icon: 'mdi:remote',
+                        unique_id: '$deviceid-remote_start_mode',
+                        state_topic: '$this/remote_start_mode',
+                        command_topic: '$this/remote_start_mode/set',
+                        name: 'Remote Start Mode',
+                        options: ['PERMANENT', 'ONE_TIME', 'OFF'],
```
anszom, line 289: These names should be more user friendly, just as you did with other enums. You may also want to move them to global constants at the top to avoid duplicating the labels.

cloud/devices/H11.ts
```diff
@@ -0,0 +1,715 @@
...
+        }
+    }
+
+    processStatistics(buf: Buffer) {
+        if (buf.length < 7) return
+
+        const sequence = buf[6]
+        if (sequence === this.lastStatSequence) {
+            // Deduplicate burst packets
+            return
```
anszom, line 587: NO need for this, `publishProperty` already deduplicates.

anszom (changes requested):
Thanks for the update. See my comments below.

PR #144, opened 2026-08-22
Add support for the LG WFV474PGV oven

jmhufford:
Adds a driver for the LG WFV474PGV double oven (deviceType 301) and wires it
into the Home Assistant bridge.

Everything was derived from live captures against a real appliance, and writes
are limited to command shapes actually observed on the wire. Fields we don't
yet understand stay visible through a diagnostic sensor rather than being
guessed at.

**Exposes:** cooktop status; per-cavity current/set temperature, state, mode,
and cook time; a settable timer; staged temperature, cook time, and Lower mode
selection; and Start and Cancel buttons per cavity.

Remote start is gated on the appliance reporting that cavity ready — both in
HA (the button goes unavailable) and again in the driver before sending.

31 tests, most built on captured frames, including both remote-start commands
reproduced byte-for-byte.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,8 @@ The following appliances are currently supported in rethink:
     - 👍 HCED3015D (STUDIO_HOOD), Generic identifier and probably works with multiple models. Working.
 - Stylers:
     - 👍 S5BBP (ST_B_E4H01Y_APL), Styler - mostly working
+- Ovens and Ranges:
+    - 🫤 WFV474PGV, Double oven/range - preliminary status, timer, cancel, and constrained remote-start support
 
 The supported appliances can be used "out of the box" with HomeAssistant or another compatible MQTT consumer.  
 Appliances not listed above can still be used with the bridge mode, but they will not be translated to MQTT. Contributions are welcome!
--- a/cloud/devices/WFV474PGV.ts
+++ b/cloud/devices/WFV474PGV.ts
@@ -1,0 +1,789 @@
+import HADevice from './base'
+import { Device as Thinq2Device } from '../thinq2/device'
+import { type Connection } from '../homeassistant'
+import { type Metadata } from '../thinq'
+import { allowExtendedType } from '@/util/casting'
+import AABBDevice from './aabb_device'
+import log from '@/util/logging'
+
+// LG oven reporting modelName "WFV474PGV" (deviceType 301, QCOM_QCA4010).
+//
+// Live captures establish the transport framing, status fields, timer commands, cavity cancel
+// selectors, and four remote-start operation profiles. Observed frame types:
+//
+//   0x40 0xEB  single current-status block, sent in reply to the status query in start()
+//   0x40 0xEC  previous and current status blocks, concatenated in that order
+//   0x40 0x31  component manifest (observed serial SAA39155903)
+//   0x40 0x72  one-shot event (timer expiry: 03 F5 0A; upper cook completion: 00 04 0A)
+//
+// Cloud-to-device writes seen so far are the F0 43 family (0x20 cavity start, 0x21 preference /
+// clock write, 0x23 timer) and the F0 44 cavity cancel.
+//
+// The known EB frame is 0x44 bytes on the wire and the EC frame is 0x82 bytes, making each status
+// block 62 bytes. Bytes that are not decoded stay out of Home Assistant; frames that are not
+// understood are logged only.
+//
+// The LG app presents three tabs: Cooktop, Upper Oven, and Lower Oven. Both oven tabs expose
+// temperature and cook time, and the app also has a settable timer. Writes are limited to captured
+// command shapes. The Upper operation exposed by the LG app is Bake; command operation 0x01 reports
+// as mode 0x81.
+//
+// The oven enforces the Remote Start arming lockout itself: a start sent to an unarmed cavity is
+// answered with the rejection acknowledgement below and nothing happens. The Start buttons are
+// gated in Home Assistant through the readiness topic's availability, and no further check is made
+// here — see the "Safety interlocks" rule in CONTRIBUTING.md.
+
+const CLASS_BYTE = 0x40
+const SINGLE_STATUS_FRAME_TYPE = 0xeb
+const STATUS_FRAME_TYPE = 0xec
+const STATUS_RECORD_LENGTH = 62
+
+// Confirmed by live off -> on and on -> off transitions on 2026-08-22. The status record contains
+// five repeated 5-byte cooktop slots beginning at offset 36. Their leading bytes all changed
+// 0x00 -> 0x01 in one EC frame when the cooktop was turned on, remained 0x01 while it stayed on, and
+// returned 0x01 -> 0x00 together in the turn-off EC frame. A later test with only the physical middle
+// burner on still made all five groups identical, confirming that these are redundant copies of a
+// global cooktop state/run time rather than independent burner states. Treat the cooktop as active
+// when any copy is active.
+const COOKTOP_SLOT_STATE_OFFSETS = [36, 41, 46, 51, 56]
+const COOKTOP_SLOT_LENGTH = 5
+
+// Confirmed by setting the app timer to 09:01:13 and observing both the initial status and the
+// 09:01:01 -> 09:00:59 rollover. Values are ordinary binary integers, not BCD.
+const TIMER_SECONDS_OFFSET = 30
+const TIMER_MINUTES_OFFSET = 31
+const TIMER_HOURS_OFFSET = 32
+
+// Preference write, subcommand 0x21, captured twice on 2026-08-29 from the LG app's Clock Settings
+// -> "Sync with smartphone". The app's Preference screen defaults every row to "No Change" and
+// cannot read any of these values back from the appliance, so one frame carries the whole set and
+// marks each untouched field 0x80. That is also what the three 0x80 bytes in the timer command are.
+// Payload layout, zero-based:
+//
+//   0      clock hours
+//   1      clock minutes
+//   2      hour format / meridiem — see CLOCK_HOUR_FORMAT_24H
+//   3      auto conversion
+//   4      temperature adjustment unit
+//   5      upper temperature adjustment, signed
+//   6      lower temperature adjustment, signed
+//   7      preheating alarm light
+//   8      beeper volume
+//   9      temperature unit shown on the oven's own display
+//   11     0xFF when untouched, a distinct sentinel from the 0x80 used everywhere else
+//   others not yet identified; every capture so far left them at 0x80
+//
+// The 0x0E length byte does not match the 13-byte payload, unlike the 0x20 start and 0x23 timer
+// subcommands where the equivalent byte does, so it is reproduced from the captures rather than
+// computed from the payload.
+const PREFERENCE_COMMAND_PREFIX = [0xf0, 0x43, 0x21, 0x0e]
+const PREFERENCE_PAYLOAD_LENGTH = 13
+const PREFERENCE_NO_CHANGE = 0x80
+const PREFERENCE_FF_INDEX = 11
+const PREFERENCE_FF_NO_CHANGE = 0xff
+const CLOCK_HOURS_INDEX = 0
+const CLOCK_MINUTES_INDEX = 1
+const CLOCK_HOUR_FORMAT_INDEX = 2
+// Byte 2 was 0x01 with the hour sent as 11, and 0x00 with the same wall-clock time sent as 23. That
+// fits both "hour format selector" and "meridiem flag", and the wire alone cannot separate them.
+// The handler writes the 0x00 shape, which carries the full 0-23 hour and is therefore unambiguous
+// under either reading; the 0x01 shape is not exposed until its meaning is confirmed.
+const CLOCK_HOUR_FORMAT_24H = 0x00
+
+// Captured one Save per value on 2026-08-29: Mute, then High, then Low, each leaving every other
+// payload byte at its no-change sentinel. Listed in the order the LG app's Beeper Volume screen
+// presents them.
+const BEEPER_VOLUME_INDEX = 8
+const BEEPER_VOLUMES = ['High', 'Low', 'Mute'] as const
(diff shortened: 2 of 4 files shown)


cloud/devices/WFV474PGV.ts
```diff
@@ -0,0 +1,544 @@
...
+                        name: 'Raw Status',
+                    },
+                    undecoded_frame: {
+                        platform: 'sensor',
+                        icon: 'mdi:help-box',
+                        entity_category: 'diagnostic',
+                        unique_id: '$deviceid-undecoded_frame',
+                        state_topic: '$this/undecoded_frame',
+                        name: 'Undecoded Frame',
+                    },
```
anszom: Please don't expose undecoded data to HA.

cloud/devices/WFV474PGV.ts
```diff
@@ -0,0 +1,544 @@
...
+            unique_id: `$deviceid-${cavity}-cancel`,
+            command_topic: `$this/${cavity}_cancel/set`,
+            payload_press: 'PRESS',
+            name: `Cancel ${title}`,
+        },
+    }
+}
+
+export default class Device extends AABBDevice {
+    // Staged remote-start parameters, keyed by appliance id. Kept outside the instance because the
```
anszom: This looks weird. Wouldn't it be easier to simply NOT publish anything in the constructor?

cloud/devices/WFV474PGV.ts
```diff
@@ -0,0 +1,544 @@
...
+        if (action === 'cancel') {
+            this.send(Buffer.from([0xf0, 0x44, CAVITY_CANCEL_SELECTORS[cavity]]))
+            return
+        }
+
+        // action === 'start'
+        if (!this.startReady[cavity]) {
+            log('WFV474PGV', `${cavity} remote start ignored because the cavity is not ready`)
+            return
+        }
```
anszom: Is this lock necessary? The `availability` trick should prevent HA from sending commands when the oven is not ready. 

Also, what would happen if we attempt to start when not ready?

cloud/devices/WFV474PGV.ts
```diff
@@ -0,0 +1,544 @@
...
+        },
+        [`${oven}_state`]: {
+            platform: 'sensor',
+            icon: 'mdi:stove',
+            unique_id: `$deviceid-${oven}_state`,
+            state_topic: `$this/${oven}_state`,
+            name: `${title} Oven State`,
+        },
+        [`${oven}_mode`]: {
+            platform: 'sensor',
```
anszom, line 233: Use `device_class: 'enum'` when the set of values is bounded. This applies to state as well.

cloud/devices/WFV474PGV.ts
```diff
@@ -0,0 +1,544 @@
...
+    static isLowerRemoteMode(value: string): value is LowerRemoteMode {
+        return (LOWER_REMOTE_MODES as readonly string[]).includes(value)
+    }
+
+    static formatOvenState(value: number) {
+        if (value === 0x00) return 'Idle'
+        if (value === 0x01) return 'Preheating'
+        if (value === 0x02) return 'Active'
+        if (value === 0x03) return 'Complete'
+        return `Unknown (0x${value.toString(16).padStart(2, '0')})`
```
anszom: Once you change device_class to `enum`, the only supported fallback is `'unknown'`.

cloud/devices/WFV474PGV.ts
```diff
@@ -0,0 +1,544 @@
...
+    }
+
+    static formatOvenMode(value: number) {
+        if (value === 0x00) return 'None'
+        return OVEN_MODE_NAMES[value & 0x7f] ?? `Unknown (0x${value.toString(16).padStart(2, '0')})`
```
anszom: See above.

cloud/devices/WFV474PGV.ts
```diff
@@ -0,0 +1,544 @@
...
+            platform: 'sensor',
+            device_class: 'temperature',
+            icon: 'mdi:thermometer-chevron-up',
+            unique_id: `$deviceid-${oven}_set_temperature`,
+            state_topic: `$this/${oven}_set_temperature`,
+            name: `${title} Oven Set Temperature`,
+            unit_of_measurement: '°F',
+        },
+        [`${oven}_cook_time`]: {
+            platform: 'sensor',
```
anszom, line 251: Use a time unit here instead of sending strings to HA.

tests/cloud/devices/WFV474PGV.test.ts
```diff
@@ -0,0 +1,766 @@
...
+
+        thinq.emit('data', BOTH_OVENS_REMOTE_START_READY)
+        assert.equal(properties.upper_oven_remote_start, 'ON')
+        assert.equal(properties.lower_oven_remote_start, 'ON')
+
+        thinq.emit('data', UPPER_OVEN_170F_THREE_MINUTES)
+        assert.equal(properties.upper_oven_state, 'Preheating')
+        assert.equal(properties.upper_oven_mode, 'Bake')
+        assert.equal(properties.lower_oven_state, 'Idle')
+        assert.equal(properties.upper_oven_temperature, 0)
```
anszom: We should probably emit `'unknown'` instead of a zero temperature. Or maybe this oven supports freezing too?:)

cloud/devices/WFV474PGV.ts
```diff
@@ -0,0 +1,544 @@
...
+            value > range.max ||
+            (value - range.min) % range.step !== 0
+        ) {
+            this.republish(prop, previous)
+            return previous
+        }
+
+        this.publishProperty(prop, value)
+        return value
+    }
```
anszom: What are we protecting against here?

README.md
```diff
@@ -43,6 +43,8 @@ The following appliances are currently supported in rethink:
     - 👍 MD19GQGE0, Smart Dehumidifier - mostly working
 - Range Hoods:
     - 👍 HCED3015D (STUDIO_HOOD), Generic identifier and probably works with multiple models. Working.
+- Ovens and Ranges:
+    - 🫤 WFV474PGV, Double oven/range - preliminary status, timer, cancel, and constrained remote-start support
```
anszom, line 49: WFV474PGV doesn't appear to be the "consumer-facing" model name.

anszom (changes requested):
Thanks for the PR! Please see my comments below.

cloud/devices/WFV474PGV.ts
jmhufford: `newDevice` builds a fresh Device on every transport reconnect, and rethink needs the staged temperature/time/mode when Start is pressed. Publishing nothing means HA keeps its retained value while rethink's copy resets to the default. Start could then send a different temperature than the one on screen. 

An alternative is publishing fixed defaults, which resets the user's selection on every reconnect.

cloud/devices/WFV474PGV.ts
jmhufford: Removed. I sent a bake command to it unarmed. it answers 40 00 43 40 instead of the usual 40 00 43 00 and does nothing.

cloud/devices/WFV474PGV.ts
jmhufford: Nothing HA can send. Removed the range checks.

tests/cloud/devices/WFV474PGV.test.ts
jmhufford: LOL. it's the oven declining to report. when I turn the oven on the current-temperature drops to 0. I updated it to publish unknown

README.md
jmhufford: I cannot find the consumer model name anywhere. oven came with the house and i cant find a serial number on it. Ill keep looking

jmhufford (commented):
Thanks for the review. I think I addressed everything

cloud/devices/WFV474PGV.ts
```diff
@@ -0,0 +1,792 @@
...
+const CAVITY_STATE_OFFSET = 0
+const CAVITY_MODE_OFFSET = 1
+const CAVITY_COOK_SECONDS_OFFSET = 2
+const CAVITY_COOK_MINUTES_OFFSET = 3
+const CAVITY_COOK_HOURS_OFFSET = 4
+const CAVITY_SET_TEMPERATURE_OFFSET = 5
+const CAVITY_CURRENT_TEMPERATURE_OFFSET = 7
+const CAVITY_REMOTE_START_OFFSET = 15
+const REMOTE_START_READY = 0x01
+const UPPER_BAKE_OPERATION = 0x01
```
anszom, line 154: From your test captures it seems that this field can take various values apart from 0x1 - maybe it's a bitfield?

cloud/devices/WFV474PGV.ts
```diff
@@ -0,0 +1,792 @@
...
+        }
+
+        if (prop === 'timer_stop' && mqttValue === 'PRESS') {
+            this.sendTimer({ hours: 0, minutes: 0, seconds: 0 })
+            return
+        }
+
+        if (prop === 'clock_sync' && mqttValue === 'PRESS') {
+            const now = new Date()
+            this.sendClock(now.getHours(), now.getMinutes())
```
anszom, line 561: This will set the oven to whatever is rethink's timezone is set to. For dockerized instances this will likely be UTC. HomeAssistant certainly has a concept of local time, but I'm not sure how we would reach it....

cloud/devices/WFV474PGV.ts
anszom: I think it's reasonable to reset the staged settings when a device reconnects and drop the map.

anszom (commented):
Just a few changes and a question about the remote start field - see below.

cloud/devices/WFV474PGV.ts
jmhufford: Done, dropped the static map. remote inits to defaults in the constructor now, so reconnect resets it.

cloud/devices/WFV474PGV.ts
jmhufford: checked the captures, value's 0x00/01/02/06/08/0E, no clean bit split. looks like a status enum. 0x01's the only value ever seen while Remote Start worked, so left it as-is and commented it.

cloud/devices/WFV474PGV.ts
jmhufford: no existing pattern for this in the repo. LG's timezone field is hardcoded and unused, HA's tz never reaches us either. Add a TZ env var to the container's setup docs?

anszom:
Thanks. I've made a few refactorings to adjust to the current `master` branch & merged. There are a few rough edges, but this has taken long enough so I wanted to get an initial version into the repo.

Take a look at the modelJSON file - you can download it through the management panel when the bridge mode is active. I can't claim that I understand the format, but it can be useful to guess. I wouldn't blindly assume that the json file is entirely accurate :)

Closing the PR now. Feel free to open another if you find any improvements are in order.

### Remote start
There is something that resembles the 'Remote start' supposed bitfield:
```
		"UpperWiFiAccess":{
			"type":"Serveral",
			"default":0,
			"option":[
				{
					"startbit":0,
					"length":1,
					"default":0,
					"superSet":"ovenState.upperRemoteStart",
					"value":"UpperRemoteStart"
				},
				{
					"startbit":1,
					"length":1,
					"default":0,
					"superSet":"ovenState.upperTimerSet",
					"value":"UpperTimerSet"
				},
				{
					"startbit":2,
					"length":1,
					"default":0,
					"superSet":"ovenState.upperOptionSet",
					"value":"UpperOptionSet"
				},
				{
					"startbit":3,
					"length":1,
					"default":0,
					"superSet":"ovenState.upperFotaEnable",
					"value":"UpperFOTAEnable"
				}
				
			]
			
		},
```

I'm not sure what is the "option set" or what influences FOTA (firmware update), but you can easily check if setting the timer affects this.

### Oven modes

Cooling is listed:

```
		"LowerOvenState":{
			"type":"Enum",
			"default":"INITIAL",
			"option":{
				"INITIAL":"@OV_STATE_INITIAL_W",
				"PREHEATING":"@OV_STATE_PREHEAT_W",
				"COOKING_IN_PROGRESS":"@OV_STATE_COOK_W",
				"DONE":"@OV_STATE_COOK_COMPLETE_W",
				"COOLING":"@OV_TERM_COOLING_W"
			}		
		},
```

### Other bitfields
"Remote Start Used" in the cooking mode?

```
		"LowerCookModeState":{
			"type":"Serveral",
			"default":0,
			"option":[
				{
					"startbit":0,
					"length":7,
					"default":0,
					"superSet":"ovenState.lowerManualCookName",
					"value":"LowerCookMode"
				},
				{
					"startbit":7,
					"length":1,
					"default":0,
					"superSet":"ovenState.lowerRemoteStartUsing",
					"value":"LowerRemoteUsing"
				}
				
			]
			
		},
```

"Temperature unit" in a 16-bit target temperature field?
```
		"LowerTargetTemp":{
			"type":"Serveral",
			"default":0,
			"option":[
				{
					"startbit":0,
					"length":15,
					"default":0,
					"superSet":"ovenState.lowerTargetTemperatureValue",
					"value":"LowerTargetTemp_F"
				},
				{
					"startbit":15,
					"length":1,
					"default":0,
					"superSet":"ovenState.lowerTargetTemperatureUnit",
					"value":"MonTempUnit"
				}
				
			]
			
		},
```

Door lock & door open states?
```
		"LowerOvenStateInfo":{
			"type":"Serveral",
			"default":0,
			"option":[
				{
					"startbit":0,
					"length":1,
					"default":0,
					"superSet":"ovenState.lowerDoorLock",
					"value":"LowerDoorlock"
				},
				{
					"startbit":1,
					"length":1,
					"default":0,
					"superSet":"ovenState.lowerControlLock",
					"value":"LowerControllock"
				},
				{
					"startbit":2,
					"length":1,
					"default":0,
					"superSet":"ovenState.lowerIsError",
					"value":"LowerError"
				},
				{
					"startbit":3,
					"length":1,
					"default":0,
					"superSet":"ovenState.lowerDoorOpen",
					"value":"LowerDoorOpen"
				}
				
			]
			
		},
```

PR #145, opened 2026-08-24
Add support for the F3P2CYUBE__ front-load washer

nickdavies:
LG WM4000HBA (deviceType 201, protocolVer 4.8, RTK_RTL8711am running clip_hna_v1.9.202).

NOTE: **This code was entirely written by Claude.** it does work with my washing machine but I don't know typescript or the repo conventions well enough to know if this is a good solution. Let me know if you want me to improve/refine this. Below is what Claude noted from 

Despite the near-identical model ID this cannot alias F3L2CYU__: that handler keys on the 0xEC/0xEB dial/status frames, and this model emits neither. Its status frame is 0xDE with a 47-byte body. The rest of the envelope matches the family -- 0x31 serial, 0xBD/0xCD full dump and idle keepalive, 0x72/0xD8 heartbeats -- so only the compact status frame differs.

Decoded conservatively: power, status, remaining_time, initial_time and a lifetime cycle counter. Offsets came from capturing a full Normal cycle and correlating byte changes against a wall-clock timeline; the remaining-time byte counted 46 -> 29 -> 16 -> 9 -> 1 -> 0 and its 16-minute reading predicted the actual finish to within 8 seconds. Unlike the F3L2CYU__ work these were NOT cross-checked against the LG cloud's decoded state (no bridge mode), so fields that could not be pinned down are left unpublished rather than guessed -- notably the course table, soil/spin/temp and error codes.

### DIFF
--- a/cloud/devices/F3P2CYUBE__.ts
+++ b/cloud/devices/F3P2CYUBE__.ts
@@ -1,0 +1,149 @@
+import HADevice from './base'
+import { Device as Thinq2Device } from '../thinq2/device'
+import { type Connection } from '../homeassistant'
+import { type Metadata } from '../thinq'
+import { allowExtendedType } from '@/util/casting'
+import AABBDevice from './aabb_device'
+
+// LG WM4000HBA front-load washer — matched on modelId "F3P2CYUBE__" (deviceType 201,
+// protocolVer 4.8, RTK_RTL8711am / clip_hna_v1.9.202). AABB frames (buf = the AABB body, AA+len and
+// checksum+BB already stripped, buf[0]==0x20 on every frame) are discriminated by buf[1]:
+//   0x31        connect/serial frame — not decoded.
+//   0xDE        status frame, 47-byte body — the only frame decoded here.
+//   0xBD / 0xCD full status dump (~443 B) / idle keepalive — only 21 bytes move across a whole
+//               cycle, so they are very mappable, but not decoded yet.
+//   0xE2        burst of ~10 frames, 2 s apart, starting the instant a cycle completes — not decoded.
+//   0x72, 0xD8  short heartbeat/ping frames — not decoded.
+//
+// NOTE this model does NOT emit the 0xEC/0xEB dial/status frames that F3L2CYU__ (WM3900HBA) keys on,
+// which is why it cannot simply alias that handler despite the near-identical model ID. Nor does it
+// report idle interactions at all: two full door open/close cycles and a dial turn, with the machine
+// powered on and holding an established MQTTS session, produced zero frames. There is consequently no
+// `door` entity here — unlike F3L2CYU__, unload has to be detected with an external sensor.
+//
+// Offsets below were derived by capturing one full Normal cycle and correlating byte changes against a
+// wall-clock timeline of observed machine behaviour. They are NOT cross-checked against the LG cloud's
+// own decoded state (bridge mode was not used), so this is weaker evidence than the F3L2CYU__ notes.
+// Fields that could not be pinned down are deliberately left unpublished rather than guessed.
+
+const STATUS_FRAME_TYPE = 0xde
+const STATUS_FRAME_LEN = 47
+
+// Minutes remaining, and the total for the current cycle. Both carry the machine's running estimate
+// while it is sensing the load, then TOTAL freezes and REMAINING counts down. Across one Normal wash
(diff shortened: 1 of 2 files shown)


cloud/devices/F3P2CYUBE__.ts
```diff
@@ -0,0 +1,149 @@
...
+// checksum+BB already stripped, buf[0]==0x20 on every frame) are discriminated by buf[1]:
+//   0x31        connect/serial frame — not decoded.
+//   0xDE        status frame, 47-byte body — the only frame decoded here.
+//   0xBD / 0xCD full status dump (~443 B) / idle keepalive — only 21 bytes move across a whole
+//               cycle, so they are very mappable, but not decoded yet.
+//   0xE2        burst of ~10 frames, 2 s apart, starting the instant a cycle completes — not decoded.
+//   0x72, 0xD8  short heartbeat/ping frames — not decoded.
+//
+// NOTE this model does NOT emit the 0xEC/0xEB dial/status frames that F3L2CYU__ (WM3900HBA) keys on,
+// which is why it cannot simply alias that handler despite the near-identical model ID. Nor does it
```
anszom, line 19: Maybe the device would be more chatty after it gets a startup request. Running in bridge mode would be helpful.

Or try this:
```
start() { this.send(Buffer.from('f0ed112101000000180413140000', 'hex')) }
```

cloud/devices/F3P2CYUBE__.ts
```diff
@@ -0,0 +1,149 @@
...
+        if (buf[1] === STATUS_FRAME_TYPE) return this.processStatus(buf)
+    }
+
+    private processStatus(buf: Buffer) {
+        if (buf.length !== STATUS_FRAME_LEN) return // reject header/layout drift
+
+        const phase = buf[PHASE_OFFSET]
+        const running = buf[RUNNING_OFFSET] === RUNNING
+
+        this.publishProperty('power', running ? 'ON' : 'OFF')
```
anszom, line 134: This entity is supposed to represent the device's power-on state. Rename the entity if you want it to report something else.

cloud/devices/F3P2CYUBE__.ts
```diff
@@ -0,0 +1,149 @@
...
+        // 'Running' is the fallback for the mid-cycle codes; a non-running frame whose phase is
+        // unrecognised reports 'Off' rather than inventing a state.
+        this.publishProperty('status', STATUS[phase] ?? (running ? 'Running' : 'Off'))
+
+        // Zeroed when not running: the machine leaves a stale minute count in these bytes after a cycle,
+        // and a countdown that keeps reading 1 while idle would be actively misleading in HA.
+        this.publishProperty('remaining_time', running ? buf[REMAINING_MIN_OFFSET] : 0)
+        this.publishProperty('initial_time', running ? buf[TOTAL_MIN_OFFSET] : 0)
+
+        this.publishProperty('cycle_count', buf[CYCLE_COUNT_OFFSET])
```
anszom, line 144: This is most likely the TCLCount field, "number of cycles since the last tub clean".

anszom (changes requested):
Thanks for the PR! Please see my comments below.

Please add a test file with real captured packets, and a README.md note.

rtvanhook:
Same model ID here (WM4000HWA, the white variant). I have a full handler up in #171: the whole 44-byte
  status record (26 fields), pause/resume/power-off/start controls, the 17 SmartCourse downloads, 23 tests
  against captured frames, README line, and a protocol writeup on my fork's wiki.

  Your 0xDE frame and my 0xEC/0xEB frames carry the same record. Your offsets 18/20/25/32 (remaining, total,
  phase, cycle count) are my rec[15]/[17]/[22]/[29] behind the same 3-byte header. The difference is
  machine state, not layout: your unit was reporting before a startup request, mine sends one in start() and
  the machine then switches to 0xEC (previous + current record) and 0xEB (single record). So the two PRs
  don't disagree about the frame.

  Would you run #171 against your HBA? No image rebuild needed, bind-mount the built device file over the
  stock dev image: https://github.com/rtvanhook/rethink/wiki/Developing-a-device-against-the-stock-image

  Two things I'd want to know from a second unit, reported on #171:
  - whether the HBA still emits 0xDE after the startup request (my handler drops that frame type today; I'll
  add it if it shows up)
  - anything that decodes wrong on your panel vs. mine

  If it checks out, this PR can close against #171, or @anszom can fold in anything he wants from it.

PR #146, opened 2026-08-24
Add LG WM3500CW (F3L2CNV4W_WIFI) front-load washer support (ThinQ1)

virtitnerd:
## Device

- Brand / model name: LG WM3500CW
- ThinQ Model ID: F3L2CNV4W_WIFI
- Platform: ThinQ1
- Related issue: none

## What's supported

Built directly from the model's own `modelJson` (fetched via `wideq`, not reverse-engineered
by hand) and confirmed against real captured commands from the physical device.

- Full read side: every field in `modelJson`'s 24-byte `Monitoring.protocol` block - status,
  timers, course/soil/spin/temp/rinse options, tub-clean count, load level, the InitialBit
  option flag, and two undecoded fields (`option3_raw`, `load_level`) exposed as diagnostic
  entities rather than guessed at.
- Control: power off, pause, and starting a cycle by course. `course_selection` only offers
  the 9 physical-dial AP courses - SmartCourse ids were tried there and removed once a real
  capture showed `OperationStart`'s SmartCourse field isn't a live selector, the machine only
  ever runs whatever's actually resident.
- Remote Start stays available through a Paused cycle, not just Off/Initial, since
  `modelJson` has no distinct Resume action - `OperationStart` is how the physical panel
  resumes a paused cycle too. `course_selection` doesn't get the same exception; there's no
  evidence it's safe to swap courses mid-pause. Remote Start also requires `remote_start`
  (RemoteStart armed on the washer itself) - this mirrors the real LG app's own restriction,
  not a lockout of our own.
- Per-cycle energy usage (`last_cycle_energy`/`course`/`completed`) via a second channel
  entirely - HTTP diagmon reports, not the persistent status socket this class otherwise
  reads from.
- Two previously-silently-discarded `thinq1/http.ts` endpoints (`diagmon` reports,
  `sendPushMessage`) now log their bodies instead of being dropped, for visibility into
  traffic that isn't decoded yet.

SmartCourse download support (a `smart_course` select that triggers a real download,
`proxyToLG()`'s byte-forwarding fix, and the new proxy/capture/synthesis machinery in
`cloud/thinq1/http.ts`) is being submitted as a separate, follow-up PR once this one lands -
per your request to split basic device support from the course-download machinery.

## How this was verified

- [x] Checked against `modelJson` (ThinQ1), not just guessed from the app's UI
- [x] Tested against a real physical device (local Docker/Portainer deployment): status,
      timers, options, course display, power off, pause, and remote start all confirmed
      working end-to-end with Home Assistant
- [x] Added `tests/cloud/devices/F3L2CNV4W_WIFI.test.ts` - 23 tests, built from real captured
      data (a mid-cycle diagmon capture, a real `OperationStart` payload)
- [x] `npm test` and `npx tsc --noEmit` pass locally

## Guidelines checklist

- [x] Maps the device protocol to Home Assistant with little or no extra logic - no timers,
      schedules, or safety locks implemented here
- [x] Doesn't circumvent any device-side safety feature
- [x] If any code here was LLM-generated, I understand it and can explain the reasoning
      behind it

### DIFF
--- a/cloud/devices/F3L2CNV4W_WIFI.ts
+++ b/cloud/devices/F3L2CNV4W_WIFI.ts
@@ -387,6 +387,11 @@ export default class Device extends HADevice {
             }),
         )
 
+        // The washer reports APCourse=0 while idle/no course actively dialed — that's not
+        // in COURSE_DEFAULTS, so the sync-on-data logic below never fires for it. Publish
+        // our own default up front so the select entity doesn't sit at "unknown" forever.
+        this.publishProperty('course_selection', AP_COURSE[this.pendingCourseId])
+
         thinq.on('data', (buf) => {
             if (buf.length < 24) return
 
--- a/cloud/devices/F3L2CNV4W_WIFI.ts
+++ b/cloud/devices/F3L2CNV4W_WIFI.ts
@@ -378,6 +378,10 @@ export default class Device extends HADevice {
         })
     }
 
+    start() {
+        this.thinq.send({ Cmd: 'Mon', CmdOpt: 'Start' })
+    }
+
     publishCache: Record<string, string | number> = {}
 
     publishProperty(prop: string, value: string | number) {
--- a/cloud/devices/F3L2CNV4W_WIFI.ts
+++ b/cloud/devices/F3L2CNV4W_WIFI.ts
@@ -27,11 +27,15 @@ import { Metadata } from '../thinq'
 // entirely: HTTP diagmon reports (cloud/thinq1/http.ts), not the persistent :47878 status
 // socket this class otherwise reads from. See the thinq.on('diagmon', ...) handler below.
 //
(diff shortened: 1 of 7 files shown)


anszom:
Thanks for the PR. It's a bit more complex than the others because of the course download machinery. Can you split this into two stages, one for the basic device support, and another for course downloads & supporting code? I suppose that the latter will require more coordination before merging.

Also, see the CONTRIBUTING.md file on the master branch. You may want to point your LLM agent to pre-review the code on your side.

virtitnerd:
> Thanks for the PR. It's a bit more complex than the others because of the course download machinery. Can you split this into two stages, one for the basic device support, and another for course downloads & supporting code? I suppose that the latter will require more coordination before merging.
> 
> Also, see the CONTRIBUTING.md file on the master branch. You may want to point your LLM agent to pre-review the code on your side.

@anszom Course Download portion has been removed. I'll open that as it's own PR once this PR gets merged in fully

PR #148, opened 2026-08-25
Add support for the LG F4W1175YW (Y_VB_Y___W.B32QEUK) washer

sakowicz:
Adds a handler for `Y_VB_Y___W.B32QEUK` — LG F4W1175YW (Vivace, EU), protocolVer 7.

Dzięki mordeczko za fajny projekt :)

### DIFF
--- a/cloud/devices/Y_VB_Y___W.B32QEUK.ts
+++ b/cloud/devices/Y_VB_Y___W.B32QEUK.ts
@@ -18,19 +18,29 @@ export default class Device extends AABBDevice {
     // consecutive frames seen with an empty selection block
     emptyStreak = 0
 
+    // pending status re-reads scheduled after a write
+    statusPollTimers: ReturnType<typeof setTimeout>[] = []
+
     constructor(HA: Connection, thinq: Thinq2Device, meta: Metadata) {
         super(HA, thinq)
         this.setConfig(
             allowExtendedType({
                 ...HADevice.config(meta, { name: 'LG Washer' }),
                 components: {
+                    // F02A0100 is a power-button press, not "set to on": sending it twice turns
+                    // the appliance on and then off again. There is also no usable feedback -
+                    // after a power change the appliance goes quiet for up to 40 seconds and
+                    // ignores F0ED status requests in the meantime, so a switch spent most of
+                    // its time showing the previous state. A button matches both the command and
+                    // what can actually be observed; `status` carries the real state once it
+                    // arrives.
                     power: {
-                        platform: 'switch',
+                        platform: 'button',
                         unique_id: '$deviceid-power',
-                        state_topic: '$this/power',
                         command_topic: '$this/power/set',
-                        name: '',
-                        icon: 'mdi:washing-machine',
+                        payload_press: '',
+                        name: 'Power',
+                        icon: 'mdi:power',
                     },
                     start: {
                         platform: 'button',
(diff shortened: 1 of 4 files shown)


cloud/devices/Y_VB_Y___W.B32QEUK.ts
```diff
@@ -0,0 +1,278 @@
...
+
+export default class Device extends AABBDevice {
+    constructor(HA: Connection, thinq: Thinq2Device, meta: Metadata) {
+        super(HA, thinq)
+        this.setConfig(
+            allowExtendedType({
+                ...HADevice.config(meta, { name: 'LG Washer' }),
+                components: {
+                    // A button, not a switch: F02A0100 toggles, and after a power change the
+                    // appliance reports nothing for up to 40s, so a switch showed a stale state.
```
anszom: maybe `optimistic: true` would suffice? Does the device report its power state correctly afterwards?

cloud/devices/Y_VB_Y___W.B32QEUK.ts
```diff
@@ -0,0 +1,278 @@
...
+                        state_topic: '$this/door_lock',
+                        name: 'Door lock',
+                        device_class: 'lock',
+                    },
+                    turbowash: {
+                        platform: 'binary_sensor',
+                        unique_id: '$deviceid-turbowash',
+                        state_topic: '$this/turbowash',
+                        name: 'TurboWash',
+                        icon: 'mdi:speedometer',
```
anszom, line 128: This device appears very similar to F_VB_F___W.B_2QEUK. Try to match icons & classes for consistency (see other entities too)

cloud/devices/Y_VB_Y___W.B32QEUK.ts
```diff
@@ -0,0 +1,278 @@
...
+        const cycles = buf[22]
+        const energy = buf[29] * 256 + buf[30]
+        const detergent = buf[31]
+        const softener = buf[32]
+        const options = buf[15]
+        const delay_end = buf[13]
+
+        // No power bit: switched off, the module stays awake and reports status=1 with the
+        // selection zeroed.
+        const powered = status > 0 && !(course === 0 && spin === 0 && temp === 0)
```
anszom, line 244: Why is this state computed? Is this coherent with the "toggle power" command?

For example: when status is nonzero, but no course is selected, does sending "toggle power" work as "On" or "Off" ? If it powers off, then - logically - the power is on prior to sending the command, and it should be reported so.

tests/cloud/devices/Y_VB_Y___W.B32QEUK.test.ts
```diff
@@ -0,0 +1,284 @@
...
+
+    test('spin and temperature hold their last value when the appliance stops reporting them', () => {
+        const { ha, thinq } = makeDevice()
+        thinq.emit('data', SAMPLE_EC_COTTON_800_60)
+        assert.equal(ha.devices[DEVICE_ID].properties.spin, 800)
+        assert.equal(ha.devices[DEVICE_ID].properties.temp, 60)
+
+        const rinsing = Buffer.from(SAMPLE_EC_COTTON_800_60)
+        const block = 2 + 14
+        rinsing[block + 1] = 7 // Rinsing
```
anszom, line 234: real captured packets would be preferable here, but I see the point of this test so I won't push too hard :)

cloud/devices/Y_VB_Y___W.B32QEUK.ts
```diff
@@ -0,0 +1,278 @@
...
+
+    start() {
+        this.send(Buffer.from('F0ED1121010000001800', 'hex'))
+    }
+
+    processAABB(buf: Buffer) {
+        if (buf[0] !== 0x20) return
+
+        // 14-byte header: buf[2..3] frame length, buf[10] record kind, buf[11..12] payload
+        // length. 0xEB carries one status block, 0xEC appends the previous one; take the first.
```
anszom, line 217: Please double-check this. `SAMPLE_EC_WASH_END_430WH` when parsed according to this rule, would indicate that the status has changed from `Off` to `End`, which seems illogical.

cloud/devices/Y_VB_Y___W.B32QEUK.ts
```diff
@@ -0,0 +1,278 @@
...
+
+        // No power bit: switched off, the module stays awake and reports status=1 with the
+        // selection zeroed.
+        const powered = status > 0 && !(course === 0 && spin === 0 && temp === 0)
+
+        this.publishProperty('error_message', ERRORS[error] ?? 'unknown') // publish message before set error state
+        this.publishProperty('error', error ? 'ON' : 'OFF')
+        this.publishProperty('status', powered ? (STATES[status] ?? 'unknown') : 'Off')
+        this.publishProperty('course', COURSES_VB[course] ?? 'unknown')
+        // temp stops being reported at Rinsing and spin at End; hold the selected value.
```
anszom, line 250: This isn't consistent with other washers. And - logically - the temperature doesn't have any meaning after the temperature-dependent part of the cycle is completed, the same holds for spin after completing the cycle.

Maybe it would be better to simply accept the `unknown`s?

anszom (changes requested):
Thanks. See my comments below.

Pozdrawiam :)

PR #153, opened 2026-08-29
RAC_056905_WW: Don't unconditionally turn off features if the device re-connects

maciejsszmigiero:
Make sure that we don't write unknown `airClean`, `jetMode` or `energySave` initial value.

Otherwise if the device disconnects from rethink then re-connects while it is running we'll always unconditionally turn each of these features OFF.

Since these devices re-connect from time to time the previous logic resulted in these features "randomly" turning off, which is definitely not what a user would expect.


anszom:
Looks good. I've replaced `== null` with `=== undefined` to be more precise.

PR #154, opened 2026-08-29
feat: add support for F14R76BSTA Front-Loading washing machine (Y_VB_Y___W.B32QEUK)

charles-huet:
Hi,

This is a WIP of the implementation of my washing machine, model F14R76BSTA.

It is missing tests, that I am in the process of writing, but so far it has been running pretty well.

There are large similarities with model F_VB_F___W_B_2QEUK, and I feel there should be a way to detect capabilities based on the device ID, but I have not had the time to look into this yet.

Best


charles-huet:
it looks like this implements the same model ID as https://github.com/anszom/rethink/pull/148

charles-huet:
@anszom This MR is ready for review and integration, as I added tests and rebased on master.

Thanks for the great work on this project !

PR #159, opened 2026-08-30
Add LG LREL6323S oven and cooktop status support

stuible:
## Summary

- add a dedicated ThinQ2 handler for the LG LREL6323S (`WLREL6323S`) electric range
- expose confirmed read-only Home Assistant entities for oven state, Bake mode, set temperature, oven door, and aggregate cooktop state
- register the model in the bridge and list its consumer-facing model name in the README

## Capture evidence

The handler and tests come from live packets captured from a physical WLREL6323S on 2026-08-30 while:

- cycling the left-front, left-rear, and right-front elements independently
- opening and closing the oven door
- starting Bake at 350°F and cancelling it

All three tested elements produced the same cooktop transition, so this exposes one aggregate Cooktop binary sensor instead of guessing at individual burner identities. Writes and uncaptured oven modes are intentionally omitted.

This is not an alias of or duplicate of #144. That PR targets the WFV474PGV double oven/range; this single-cavity model has different observed door and cooktop field layouts and therefore needs its own handler.

## Verification

- `npm run build`
- `node --import tsx --test --test-reporter=spec tests/cloud/devices/WLREL6323S.test.ts` (8/8 pass)
- full repository suite: 444 tests passed in the sandbox; the 7 loopback-binding tests blocked there were rerun outside the sandbox and all 7 passed
- `npx prettier . --check --ignore-path .gitignore`
- `git diff --check`

### DIFF
--- a/cloud/devices/WLREL6323S.ts
+++ b/cloud/devices/WLREL6323S.ts
@@ -1,0 +1,157 @@
+import HADevice from './base'
+import { Device as Thinq2Device } from '../thinq2/device'
+import { type Connection } from '../homeassistant'
+import { type Metadata } from '../thinq'
+import { allowExtendedType } from '@/util/casting'
+import AABBDevice from './aabb_device'
+import log from '@/util/logging'
+
+// LG LREL6323S electric range, reporting modelName "WLREL6323S" (deviceType 301).
+//
+// Captures from the physical panel establish the single-cavity oven fields, door-open bit,
+// and aggregate cooktop state. Normal elements and the warming zone use different bytes, but
+// individual normal elements report the same transition, so no individual entities are exposed.
+// Writes remain disabled until their complete command shapes and appliance interlocks are captured.
+
+const CLASS_BYTE = 0x40
+const SINGLE_STATUS_FRAME_TYPE = 0xeb
+const STATUS_FRAME_TYPE = 0xec
+const STATUS_RECORD_LENGTH = 62
+
+const OVEN_STATE_OFFSET = 0
+const OVEN_MODE_OFFSET = 1
+const OVEN_SET_TEMPERATURE_OFFSET = 5
+const OVEN_DOOR_FLAGS_OFFSET = 11
+const OVEN_DOOR_OPEN_MASK = 0x08
+// Normal surface elements set bytes 36 and 38 together; the warming zone sets
+// bytes 37 and 39 together. Treat all four as one read-only cooktop status.
+const COOKTOP_STATE_OFFSETS = [36, 37, 38, 39]
+
+const OVEN_STATE_NAMES: Record<number, string> = {
+    0x00: 'Idle',
+    0x01: 'Preheating',
+    0x02: 'Cooking',
+    0x04: 'Cooling',
+    0x09: 'Cleaning',
+}
+const OVEN_MODE_NAMES: Record<number, string> = {
+    0x00: 'None',
+    0x01: 'Bake',
+    0x03: 'Convection Bake',
+    0x04: 'Convection Roast',
+    0x07: 'Broil',
+    0x08: 'Warm',
+    0x0d: 'EasyClean',
+    0x0f: 'Self Clean',
+    0x10: 'Air Fry',
+}
+const OVEN_STATE_OPTIONS = Object.values(OVEN_STATE_NAMES)
+const OVEN_MODE_OPTIONS = Object.values(OVEN_MODE_NAMES)
+
+// Observed from LG's cloud and confirmed against the WLREL6323S: the reply is a 40 EB status frame.
+const STATUS_QUERY = 'f0ed114101000000181a0207080c14191a1e262b30353a00000000000000000000000000'
+
+export default class Device extends AABBDevice {
+    constructor(HA: Connection, thinq: Thinq2Device, meta: Metadata) {
+        super(HA, thinq)
+        this.setConfig(
+            allowExtendedType({
+                ...HADevice.config(meta, { name: 'LG Electric Range' }),
+                components: {
+                    cooktop_status: {
+                        platform: 'binary_sensor',
+                        icon: 'mdi:stove',
+                        unique_id: '$deviceid-cooktop_status',
+                        state_topic: '$this/cooktop_status',
+                        name: 'Cooktop',
+                        payload_on: 'ON',
+                        payload_off: 'OFF',
+                    },
+                    oven_status: {
+                        platform: 'sensor',
+                        device_class: 'enum',
+                        icon: 'mdi:stove',
+                        unique_id: '$deviceid-oven_status',
+                        state_topic: '$this/oven_status',
+                        name: 'Oven Status',
+                        options: OVEN_STATE_OPTIONS,
+                    },
+                    oven_mode: {
+                        platform: 'sensor',
+                        device_class: 'enum',
+                        icon: 'mdi:chef-hat',
+                        unique_id: '$deviceid-oven_mode',
+                        state_topic: '$this/oven_mode',
+                        name: 'Oven Mode',
+                        options: OVEN_MODE_OPTIONS,
+                    },
+                    oven_set_temperature: {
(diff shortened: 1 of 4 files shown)


cloud/devices/WLREL6323S.ts
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    static formatOvenState(value: number) {
+        const state = OVEN_STATE_NAMES[value]
+        if (state === undefined) log('WLREL6323S', 'undecoded oven state', value.toString(16))
+        return state ?? 'unknown'
+    }
+
+    static formatOvenMode(value: number) {
+        const mode = OVEN_MODE_NAMES[value & 0x7f]
```
anszom: Why the mask? What is on the high bit?

The modelJSON file suggests that this is some kind of a remote-start flag.

cloud/devices/WLREL6323S.ts
```diff
@@ -0,0 +1,146 @@
...
+const OVEN_STATE_OFFSET = 0
+const OVEN_MODE_OFFSET = 1
+const OVEN_SET_TEMPERATURE_OFFSET = 5
+const OVEN_DOOR_FLAGS_OFFSET = 11
+const OVEN_DOOR_OPEN_MASK = 0x08
+const COOKTOP_STATE_OFFSETS = [36, 38]
+
+const OVEN_STATE_NAMES: Record<number, string> = {
+    0x00: 'Idle',
+    0x01: 'Preheating',
```
anszom, line 32: This seems a bit incomplete, same for the mode names list. You can probe them by sending altered packets to the LG cloud via the bridge interface, or read them out from the modelJSON file.

cloud/devices/WLREL6323S.ts
```diff
@@ -0,0 +1,146 @@
...
+const SINGLE_STATUS_FRAME_TYPE = 0xeb
+const STATUS_FRAME_TYPE = 0xec
+const STATUS_RECORD_LENGTH = 62
+
+const OVEN_STATE_OFFSET = 0
+const OVEN_MODE_OFFSET = 1
+const OVEN_SET_TEMPERATURE_OFFSET = 5
+const OVEN_DOOR_FLAGS_OFFSET = 11
+const OVEN_DOOR_OPEN_MASK = 0x08
+const COOKTOP_STATE_OFFSETS = [36, 38]
```
anszom: Why two offsets? This may be ok, but no rationale was provided.

cloud/devices/WLREL6323S.ts
```diff
@@ -0,0 +1,146 @@
...
+                    cooktop_status: {
+                        platform: 'binary_sensor',
+                        icon: 'mdi:stove',
+                        unique_id: '$deviceid-cooktop_status',
+                        state_topic: '$this/cooktop_status',
+                        name: 'Cooktop',
+                        payload_on: 'ON',
+                        payload_off: 'OFF',
+                    },
+                    oven_state: {
```
anszom: We use 'status' in most (all?) other devices, so 'oven_status' would be more consistent.

cloud/devices/WLREL6323S.ts
```diff
@@ -0,0 +1,146 @@
...
+                    oven_mode: {
+                        platform: 'sensor',
+                        device_class: 'enum',
+                        icon: 'mdi:chef-hat',
+                        unique_id: '$deviceid-oven_mode',
+                        state_topic: '$this/oven_mode',
+                        name: 'Oven Mode',
+                        options: OVEN_MODE_OPTIONS,
+                    },
+                    oven_set_temperature: {
```
anszom: There is no merged oven yet, but #165 uses 'target_temperature' so maybe you should follow that scheme here as well.

cloud/devices/WLREL6323S.ts
```diff
@@ -0,0 +1,146 @@
...
+                        icon: 'mdi:thermometer-chevron-up',
+                        unique_id: '$deviceid-oven_set_temperature',
+                        state_topic: '$this/oven_set_temperature',
+                        name: 'Oven Set Temperature',
+                        unit_of_measurement: '°F',
+                    },
+                    oven_door: {
+                        platform: 'binary_sensor',
+                        device_class: 'door',
+                        icon: 'mdi:door',
```
anszom: leaving the icon unset will (allegedly) allow the icon to automatically toggle based on the door state.

tests/cloud/devices/WLREL6323S.test.ts
```diff
@@ -0,0 +1,135 @@
...
+    test('is registered in the ThinQ2 bridge', () => {
+        const ha = new MockHAConnection()
+        const thinq = new MockThinq2Device(DEVICE_ID, META)
+        const bridge = new Bridge(ha.asConnection())
+        bridge.newDevice(thinq)
+
+        assert.ok(ha.devices[DEVICE_ID].config)
+        assert.equal(bridge.haDevices.has(DEVICE_ID), true)
+        assert.equal(thinq.outbox.length, 1)
+    })
```
anszom: The test should be focused on the protocol decoder. Bridging is not really affected by anything specific to this device, so this test can be removed.

anszom (changes requested):
Thanks for the PR, see my comments below.

anszom:
(I've deleted a comment about renaming the set-temperature entity, if you've managed to notice it, please ignore it:)

cloud/devices/WLREL6323S.ts
stuible: Expanded both maps from new physical-panel captures. Status values now cover Idle 0x00, Preheating 0x01, Cooking 0x02, Cooling 0x04, and Cleaning 0x09. Mode values now cover every function listed for this model and exercised on the appliance: Bake 0x01, Convection Bake 0x03, Convection Roast 0x04, Broil 0x07, Warm 0x08, EasyClean 0x0d, Self Clean 0x0f, and Air Fry 0x10. The tests use the real captured EC frames for each mode.

cloud/devices/WLREL6323S.ts
stuible: Renamed the entity and published property to oven_status for consistency.

cloud/devices/WLREL6323S.ts
stuible: New captures resolved this: ordinary surface elements set bytes 36 and 38 together, while the center-rear warming zone sets bytes 37 and 39 together. The aggregate sensor now checks all four bytes, with a real warming-zone fixture and rationale documented in the handler.

cloud/devices/WLREL6323S.ts
stuible: Removed the explicit door icon so Home Assistant can select the state-aware icon from device_class: door.

tests/cloud/devices/WLREL6323S.test.ts
stuible: Removed the bridge-registration test and its Bridge import. The suite is now focused on the device protocol decoder and real captured packets.

cloud/devices/WLREL6323S.ts
stuible: Removed the mask. Live panel captures showed that enabling Remote Start changes record byte 15 from 0x0e to 0x01 while the oven-mode byte remains 0x00; disabling it reverses the same transition. No high bit was observed on the mode byte, so unknown values (including 0x81) now remain unknown instead of being mislabeled as Bake.

anszom: the exact name is "RemoteStartUsing" -> maybe the flag is set when the oven is started via remote start? That's just my guess.

anszom: I've made the same comment in PR #144, so the conclusion may apply to both.

anszom:
Thanks. I've rebased this onto master & refactored to use the new `Enum` utility. Closing the PR - although I'd ask you to take a look at the remote start bit mentioned in a comment above. If you find anything to be improved, feel free to open another PR.

Issue #93, opened 2026-07-14
Protocol: WMVEL2137 microwave/hood — fan & light control + state encoding


XanderLuciano:
## Device

- **Hardware model:** MVEL2033F (LG over-the-range microwave/hood combo)
- **WiFi-reported model:** WMVEL2137
- **Device Type:** 302
- **Chipset:** Realtek RTL8720cm (CLIP BLE v1.9.210)
- **Protocol:** ThinQ2, Service Code SVC202

## Control Commands (Cmd 04)

Sent on `lime/devices/<deviceId>` MQTT topic, JSON-wrapped:
```json
{"did":"<id>","mid":<ts>,"cmd":"packet","type":1,"data":"<hex>"}
```

The command format uses advance-count semantics — there is no direct level setting. Levels are reached by specifying how many presses to advance from current state:

| Target | Advances from OFF |
|---|---|
| Fan LOW | 1 |
| Fan MEDIUM | 2 |
| Fan HIGH | 3 |
| Fan TURBO | 4 |
| Light LOW | 1 |
| Light HIGH | 2 |

**Advances are relative to current state** and wrap around (e.g., advancing 2 from HIGH: HIGH→OFF→LOW). To reliably hit a target level, advance from OFF or track current state.

### Packet Structure
```
aa 0e f0 43  22 04  [FAN     ][LIGHT   ][MTR/TIME] [chk] bb
│  │  └──┬──┘  └─┬─┘  │  │     │  │     │   │       │    └ end
│  │     │       │    │  │     │  │     │   │       └ checksum (see below)
│  │     │       │    │  │     │  └──adv  │   └ time (0x80=no-change)
│  │     │       │    │  │     └─cmd      └ motor (0x80=no-change)
│  │     │       │    │  └─advance count  (fan & light)
│  │     │       │    └─cmd: 01=advance  00=off  80=no-change  (light)
│  │     │       └ sub-command (0x2204)
│  │     └ message type (0xF043)
│  └ payload length (14 bytes)
└ start byte
```

**Field order is FAN first, then LIGHT** — matches the `53XX` status nibble order. The stove confirms this: when it turns on, it sends `01 02 01 02` which produces light HIGH + fan MED (not the other way around).

### Checksum
```
checksum = (sum of [AA, LEN, BODY...] mod 256) XOR 0x55
```

### Command Reference
| Command | Hex | Notes |
|---|---|---|
[...]

PR #160, opened 2026-08-30
Add LG MVEM1825 microwave support

stuible:
## Summary

- add a ThinQ2 handler for the WMVEM1825 protocol used by MVEM1825D/F over-the-range microwaves
- expose cooking status, remaining time, power level, and door state
- expose controllable two-level vent fan and cooktop light entities
- preserve the other hood control when either fan or light is changed
- register the model and document the marketing model names

## Hardware validation

All status fixtures come from the live device and were matched to physical actions. Captures cover idle, ready, cooking, paused, done, 10-second and 1:05 countdowns, power levels 5 and 10, door open/closed, and both hood levels.

ThinQ app commands were captured separately for fan and light. The app explicitly clears the other hood control when changing one. A combined fan High + light Low command was therefore tested directly, acknowledged by the appliance, reflected in its status packet, and physically confirmed before implementing state preservation.

This is intentionally a separate handler from the WMVEL2137 protocol documented in #93: WMVEM1825 has two direct fan levels and packs light in the upper status nibble, while that device reports four advance-count fan levels with the opposite nibble layout.

## Tests

- npm test
- npm run build
- 14 WMVEM1825 handler tests using real captured packets

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,8 @@ The following appliances are currently supported in rethink:
     - 👍 MD19GQGE0, Smart Dehumidifier - mostly working
 - Range Hoods:
     - 👍 HCED3015D (STUDIO_HOOD), Generic identifier and probably works with multiple models. Working.
+- Microwave Ovens:
+    - 👍 MVEM1825D/F (WMVEM1825), Smart Over-the-Range Microwave - mostly working
 - Stylers:
     - 👍 S5BBP (ST_B_E4H01Y_APL), Styler - mostly working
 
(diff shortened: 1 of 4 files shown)


anszom (commented):
Thanks. This looks good overall. The master branch has diverged somewhat but I can reconcile this at merge time. One question though:

My LLM agent has pointed out that rec[10] is equal to rec[20] in all of your captures. The latter appears to be a part of the 'total time' sequence on bytes 18..20. Are you sure that the power level is properly decoded? A single capture with a cooking level not equal to the total number of seconds would clarify this.

stuible:
Confirmed with a new physical capture: a 20-second cook at power level 5 reports rec[10] = 5, rec[17] = 20 remaining seconds, and rec[20] = 20 target seconds. I added that exact packet as a regression test, amended the original commit, and the CI checks for Node 20 and Node 24 are green.

anszom:
Ok, merged (+refactor). Closing the PR. Thanks:)

PR #162, opened 2026-09-05
Add support for the LG MVEL2033F over-the-range microwave

mattst88:
Adds a handler for modelId WMVEL2137 (thinq2 deviceType 302), the
over-the-range microwave and hood sold as MVEL2033F. Read and write,
verified on the appliance.

This is a different appliance from WMVEM1825 and cannot be aliased to
it: that model has two fan levels, this one four plus off, and the two
pack the vent and lamp nibbles in opposite halves of their status byte.

Published read-only: the appliance state, the cook mode, the dish being
cooked, the power level, the remaining time, the length the cook was
given, and the kitchen timer. The vent fan is a fan platform and the
lamp a light platform, both writable.

Confirmed against live capture, with rethink bridged to the LG cloud so
every byte change could be matched against LG's own decoded state at
the same timestamp:

- The five cook modes, each against LWOManualCookName.
- rec[6:8] as a big-endian dish id, reported verbatim by the cloud as
  LWOSubCookName. 25 dishes are named from the panel.
- The vent and lamp nibbles of rec[36], nine consecutive single-variable
  transitions for the fan and four for the lamp. Issue #93 documents the
  opposite nibble order for this modelId; on this appliance it is
  unambiguous.
- The write path: a 10-byte inner body, with the level absolute rather
  than an advance count, which is the other correction to issue #93.
  Sending level 2 while the fan ran at 4 gave 2, and "advance 4" from
  Low gave 4 rather than wrapping to off.
- All three durations, on cooks of 30 seconds, five and 90 minutes and
  timers of one, 25, 61 and 90 minutes.

The two countdowns are a minutes byte and a seconds byte with no hour
in front, despite the layout suggesting one: a 90-minute timer reads 90
and a 61-minute one counts through 60:59 to 59:58 without borrowing,
and the cloud reports LWOTimerMinute:90 rather than an hour and thirty.
A byte holds 255 minutes, past anything either field needs.

The cook length at rec[18:20] is the one field that does split into
hours and minutes, matching LWOTargetTimeHour and LWOTargetTimeMinute:
a 90-minute cook reads 1 and 30 there while the remaining time counts a
flat 90 down. It holds steady while the remaining time ticks, and both
bytes clear when the cook is cancelled. It carries no seconds, so a
cook shorter than a minute reports 0, which is what LG's own field
does.

Outstanding and unknown:

- rec[15] and rec[21], the bytes ahead of each countdown pair, have been
  0 in every record ever captured and are deliberately not read: taking
  them for hours, which their position suggests, would silently add
  whole hours to a duration if they are something else.
- Dish ids are sparse and only those 25 have been read off the panel.
  An unlisted id reports unknown rather than a guess. The alphabetical
  ordering that corroborates the Sensor Cook, Melt and Steam Cook blocks
  does not extend to Defrost or the Kids Menu, so those entries rest on
  the panel reading alone.
- The modelJSON carries a 12-entry error table that has not been
  exercised, since that needs a real fault. No error entity is exposed.
- The vent can change level with no command from rethink: with the
  range-hood interaction enabled in the ThinQ app, any cooktop element
  on the matching LG range brings the fan to level 1. Documented at
  VENT_LAMP_OFFSET, since a driver assuming the level only moves on
  command would be wrong.

### DIFF
--- a/cloud/devices/WMVEL2137.ts
+++ b/cloud/devices/WMVEL2137.ts
@@ -1,0 +1,392 @@
+import HADevice from './base'
+import { Device as Thinq2Device } from '../thinq2/device'
+import { type Connection } from '../homeassistant'
+import { type Metadata } from '../thinq'
+import { allowExtendedType } from '@/util/casting'
+import AABBDevice from './aabb_device'
+import { Enum } from '@/util/enum'
+
+// LG over-the-range microwave/hood, marketing model MVEL2033F, matched on modelId "WMVEL2137"
+// (thinq2 deviceType 302).
+//
+// A different appliance from WMVEM1825 and not aliasable to it: that model has two fan levels
+// against this one's four plus off. Both pack the vent/lamp status nibbles the same way (see
+// VENT_LAMP_OFFSET) - it is issue #93's claim for this modelId that is reversed, not WMVEM1825.
+//
+// Frames arrive 0xFF-escaped AABB. The AABB body seen by processAABB is:
+//
+//   41 EC <record: previous state> <record: current state>     94 bytes
+//   41 EB <record: current state>                               48 bytes
+//
+// Each record is 46 bytes. 0xEB answers the status query sent from start(); 0xEC is the
+// unsolicited delta report, whose first record repeats the previous frame's current state. Both
+// frame types share one decoder over the current-state record.
+//
+// Every offset and table entry was confirmed live: the appliance was driven by hand at the panel
+// while rethink bridged to the LG cloud, matching each byte change to the cloud's own ovenState
+// at the same timestamp.
+//
+// Credit: the protocol, the AABB checksum and the fan/light command shape were first documented
+// by XanderLuciano in anszom/rethink issue #93. Two of that issue's claims do not hold here: the
+// vent/lamp nibble order is reversed (VENT_LAMP_OFFSET), and the level byte is absolute rather
+// than an advance count (SET_COMMAND). The state and cook-mode codes also are not the modelJSON
+// enum indices.
+
+const STATUS_FRAME_TYPE = 0xec
+const STATUS_FRAME_LEN = 94 // 2B header + 46B previous record + 46B current record
+const CURRENT_RECORD_OFFSET = 48
+
+const SINGLE_STATUS_FRAME_TYPE = 0xeb
+const SINGLE_STATUS_FRAME_LEN = 48 // 2B header + 46B record, no preceding "previous state" half
+const SINGLE_RECORD_OFFSET = 2
+
+const RECORD_LEN = 46
+const CLASS_BYTE = 0x41
+
+// Status query, sent on connect. Without it the appliance pushes nothing until the panel is
+// touched. Bytes as sent by LG's own cloud while bridged, so known to be read-only and understood.
+const STATUS_REQUEST = 'f0ed114101000000180e111718191a1a1b00000000000000'
+
+// rec[0]: appliance state, matched against the cloud's LWOState. Not the modelJSON UpperOvenState
+// enum index (5 is DONE here, CLEANING there; 4 is PAUSED here, COOLING there).
+const STATE_OFFSET = 0
+const STATES = Enum.of({
+    Idle: 0x00,
+    Cooking: 0x02,
+    Paused: 0x04,
+    Done: 0x05,
+    Settings: 0x07,
+})
+
+// rec[5]: cook mode family, matched against LWOManualCookName. Not the enum index either (SENSOR_
+// COOK is 0x18 here, index 17 there). 0x18 covers every Sensor Cook item, 0x1f every preset menu
+// item (Kids Meal, Steam Cook, Melt, Soften all report it) - these name the family, not the dish.
+const COOK_MODE_OFFSET = 5
+const COOK_MODES = Enum.of({
+    Off: 0x00,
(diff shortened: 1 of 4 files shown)


cloud/devices/WMVEL2137.ts
```diff
@@ -0,0 +1,436 @@
...
+// and the cloud reports LWOTimerMinute:90 rather than an hour and thirty. A byte holds 255 minutes,
+// comfortably past anything either field needs.
+//
+// That leaves rec[15] and rec[21], the bytes ahead of each pair, undecoded. Both have been 0 in
+// every record ever captured. They are deliberately not read: taking them for hours, which is what
+// their position suggests, would silently add whole hours to a duration if they turn out to be
+// something else.
+const REMAIN_TIME_OFFSET = 16
+const TIMER_OFFSET = 22
+// rec[18..19]: the length the running cook was given, as hours then minutes with no seconds byte,
```
anszom: Are you sure `rec[20]` is not the seconds counter? For example, `SAMPLE_COOKING_HALF_POWER_30S` appears to have the right value there.

cloud/devices/WMVEL2137.ts
```diff
@@ -0,0 +1,436 @@
...
+import { type Metadata } from '../thinq'
+import { allowExtendedType } from '@/util/casting'
+import AABBDevice from './aabb_device'
+
+// LG over-the-range microwave/hood, marketing model MVEL2033F, matched on modelId "WMVEL2137"
+// (thinq2 deviceType 302, Realtek RTL8720cm running CLIP BLE v1.9.210).
+//
+// This is a different appliance from WMVEM1825 and needs its own handler: that model has two
+// direct fan levels, this one has four plus off, and the two pack the vent/lamp status nibbles
+// in opposite halves of the byte (see VENT_LAMP_OFFSET).
```
anszom: PR #160 (WMVEM1825) has the same nibble order. The rest of the comment is justified though.

cloud/devices/WMVEL2137.ts
```diff
@@ -0,0 +1,436 @@
...
+//     cmd:   0x01 set this control to the level that follows
+//            0x00 turn this control off
+//            0x80 leave this control exactly as it is
+//     level: ABSOLUTE - fan 0-4, lamp 0-2. Issue #93 describes this byte as a count of presses
+//            to advance by; on this appliance it is not. Sending fan level 2 while the fan was
+//            running at 4 gave 4 -> 2, and "advance 4" from Low gave 1 -> 4 rather than wrapping
+//            round to off. The advance reading happens to agree only when starting from off.
+//     trailer: 0x80, "no change", covering the motor/time fields issue #93 documents.
+//
+// The shape is LG's own, taken from watching the ThinQ app drive this appliance through all four
```
anszom: Can you reduce the volume of the comments? It's a bit too much.

anszom (commented):
Looks good, just some minor fixes.

Please also take a look at entity names/icons and check if they can be better aligned with other similar devices. Also compare with #160 - but that one hasn't been merged yet - so there is no clear priority here.

While you're at it, can you rebase this on top of the current master and make use of the `Enum` class?

cloud/devices/WMVEL2137.ts
mattst88: Yes, sorry.

cloud/devices/WMVEL2137.ts
mattst88: I'll double check.

mattst88: Confirmed and fixed: rec[20] is a seconds byte. cook_time was truncating sub-minute lengths to 0 (SAMPLE_COOKING_HALF_POWER_30S had cook_time=0 with nothing catching it). Now reads hours/minutes/seconds, with a regression test on that sample.

cloud/devices/WMVEL2137.ts
mattst88: Trimmed the comments down.

cloud/devices/WMVEL2137.ts
mattst88: Ack, thanks.

cloud/devices/WMVEL2137.ts
mattst88: Yep! You're right: double-checked against three more live samples in the existing bridged capture, all non-round-minute cooks so rec[20] isn't masked by being 0:

- LWOTargetTimeMinute:1, LWOTargetTimeSecond:10 → rec[18..20] = 0,1,10
- LWOTargetTimeMinute:3, LWOTargetTimeSecond:20 → rec[18..20] = 0,3,20
- LWOTargetTimeMinute:1, LWOTargetTimeSecond:40 → rec[18..20] = 0,1,40

Fixed.

cloud/devices/WMVEL2137.ts
mattst88: Actually this needed a real fix, not just an ack — the file comment claimed WMVEM1825 packs the nibbles the opposite way from this appliance, but #160's code is `fanSpeed = hood & 0x0f; lightLevel = hood >> 4`, same low/high order as here. Fixed the comment to say the two appliances agree and it's issue #93's claim for this modelId that's reversed, not WMVEM1825.

anszom:
Thanks. Merged with some small adjustments. Closing the PR.

PR #163, opened 2026-09-05
Add support for the LG WM5800HVA front-load washer

mattst88:
Adds a handler for modelId FAFXU25006 (thinq2 deviceType 201), the
front-load washer sold as WM5800HVA. Read-only.

Frames are the 0xFF-escaped AABB form with 72-byte records, class byte
0x20, against 42 on the matching dryer and 28 on the RV13* models, with
the fields at different offsets from either, so this cannot be aliased.

Published: the cycle state and the state before it, the course, soil
level, temperature, rinse and spin settings, the remaining time, the
cycle length, the delay time, the energy the cycle has used, the load
level, the buzzer volume, the washes since the last Tub Clean, both
EZDispense settings, the door lock, the door itself, and the Turbo
Wash, Pre Wash, Auto Soak, Steam, Fresh Care, Child Lock, drum light,
Delay Wash, Remote Maintain and end melody flags.

Confirmed against live capture with rethink bridged to the LG cloud, so
each byte change could be matched against LG's own decoded state at the
same timestamp. Of note:

- The minute counters are little-endian here, the opposite of the
  matching dryer's.
- This model's modelJSON enum indices are the wire codes, unlike the LG
  range and microwave, so the full state enum is carried. Nine states
  were seen live, eight of them against the cloud's own state field.
- Course has no index table in the modelJSON at all, so all 29 codes
  came from selecting each cycle and reading the byte. Fifteen are dial
  positions and the rest are reachable only from the ThinQ app. The
  cloud's panelCrsList is not a catalogue and should not be mistaken
  for one: it is the ten slots currently assigned to the panel, and two
  versions captured an hour apart differed in six of the ten entries.
  Several cycles never appeared in either, so the table is very likely
  still incomplete.
- The course NAMES are read off the panel rather than taken from LG's
  cloud identifiers, which disagree on eight of the fifteen dial
  positions. Some of that is wording, SPEEDWASH being Quick Wash and
  RINSE_SPIN being "Rinse & Spin", but some would mislead outright:
  DUVET is the cycle the panel calls Bedding, and ALLERGYCARE and WHITE
  are LG's branded Allergiene and BrightWhites. Each entry keeps its
  identifier as a trailing comment, since that is what the cloud and a
  capture will show. The trademark symbols the panel prints on
  Allergiene, ColdWash and BrightWhites are dropped as decoration.
- A Delay Wash was followed from armed to firing: rec[12:14] holds the
  value while the machine waits in INITIAL and counts down only once it
  reaches RESERVED.
- The two EZDispense bytes are told apart by the phase each returns to
  LG's dash sentinel in, detergent as the wash ends and softener as the
  rinse does, the latter named by the cloud in that frame.
- rec[28] is TCLCount, stepping by one in the frame each cycle reaches
  END.

The door is the one published field with no cloud pairing, and that is
a limit of the cloud rather than of the evidence. It is not in the
status record at all, but announced in a separate 0x03 event frame.
LG's decoded state has a doorClose field, but it is never sent in a
change notification: across 43 door movements it appeared in none, only
in the periodic full dumps, and in all six of those it read
DOORCLOSE_OFF including two taken while the door was open. So no amount
of driving the appliance will pair it, and the wire is the only source
that reports this door. The byte itself was isolated by ten alternating
transitions plus a pair driven one action at a time, and the EZDispense
drawer emits no such frame in either direction.

Outstanding and unknown:

- The end melody at rec[69] has never moved independently of the
  buzzer, so its isolation rests on LG's field name rather than on a
  single-variable change.
- No dispenser level above 0 has been seen. Less, Normal and More come
  from LG's enum for the field, not from observation.
- rec[37] bits 0x02 and 0x10 match no cloud field: 0x02 tracks the panel
  being awake and 0x10 a live cycle. Both are documented, neither is
  published.
- The error code was never located; that needs a real fault. Nor was any
  cycle-progress byte of the kind the matching dryer has.
- The panel powers itself off after a couple of idle minutes, clearing
  the course, times and settings to their dash sentinels. Worth knowing
  before reading anything into a record that has gone blank.

### DIFF
--- a/cloud/devices/FAFXU25006.ts
+++ b/cloud/devices/FAFXU25006.ts
@@ -1,0 +1,570 @@
+import HADevice from './base'
+import { Device as Thinq2Device } from '../thinq2/device'
+import { type Connection } from '../homeassistant'
+import { type Metadata } from '../thinq'
+import { allowExtendedType } from '@/util/casting'
+import AABBDevice from './aabb_device'
+import { Enum } from '@/util/enum'
+
+// LG front-load washer, matched on modelId "FAFXU25006" (thinq2 deviceType 201), sold as WM5800HVA.
+//
+// This model uses the long, length-escaped AABB framing (aa ff <class> <envelope> <len16> ... <inner
+// type> <inner len> <payload> <chk16> bb) rather than the short one every other AABB handler here
+// reads. AABBDevice.processData strips AA/length and the trailing BB, so buf[0] is the class byte and
+// the inner type sits at buf[10]. The checksum is 16-bit here, so buf still carries one checksum byte
+// past the payload, which is why the expected buf lengths below are one more than header + inner len.
+//
+// Inner types:
+//   0xEC  two stacked 72-byte records, previous state at buf[13] and current state at buf[85]
+//   0xEB  a single current-state record at buf[13], same layout, sent in reply to a status query
+//   0x03  a 69-byte event frame sent when the door opens or shuts — not a status record
+//   0x00/0x02/0x04/0x4d/0xe2  not status records and not decoded
+//
+// Record length is 72 bytes here against 42 on the LG dryers and 28 on the RV13* reference, at
+// completely different offsets, so this cannot be aliased to any existing handler.
+//
+// Every offset below is live-verified: real wire traffic captured while the washer was driven by hand
+// at the panel, with rethink bridged to the LG cloud so each byte change could be matched against the
+// cloud's own decoded washerDryer state at the same timestamp.
+
+const CLASS_BYTE = 0x20
+const ENVELOPE_TYPE = 0x0a
+const INNER_TYPE_OFFSET = 10
+const INNER_LEN_OFFSET = 11
+
+const RECORD_LEN = 72
+const RECORD_MARKER = 0x0a // the current-state record's own marker byte, rec[0]
+
+const STATUS_FRAME_TYPE = 0xec
+const STATUS_INNER_LEN = 2 * RECORD_LEN
+const STATUS_RECORD_OFFSET = 13 + RECORD_LEN // skip the "previous state" record
+const STATUS_BUF_LEN = 13 + STATUS_INNER_LEN + 1
+
+const SINGLE_STATUS_FRAME_TYPE = 0xeb
+const SINGLE_INNER_LEN = RECORD_LEN
+const SINGLE_RECORD_OFFSET = 13
+const SINGLE_BUF_LEN = 13 + SINGLE_INNER_LEN + 1
+
+// The door does not live in the status record: opening or shutting it moves nothing in the 72 bytes.
+// It is announced in its own frame, inner type 0x03, with buf[22] reading 1 for open and 2 for shut.
+// LG's cloud doorClose field is never sent in a change notification and reads DOORCLOSE_OFF even while
+// open in the periodic dumps, so this is the one published field here with no cloud pairing behind it —
+// the wire is the only source that reports this door.
+const DOOR_EVENT_TYPE = 0x03
+const DOOR_EVENT_INNER_LEN = 14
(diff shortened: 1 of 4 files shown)


cloud/devices/FAFXU25006.ts
```diff
@@ -0,0 +1,674 @@
...
+                    spent_power: {
+                        platform: 'sensor',
+                        unique_id: '$deviceid-spent_power',
+                        state_topic: '$this/spent_power',
+                        name: 'Cycle energy',
+                        icon: 'mdi:lightning-bolt',
+                        device_class: 'energy',
+                        unit_of_measurement: 'Wh',
+                        // resets to 0 with each new cycle, so it is a per-cycle total, not a meter
+                        state_class: 'total',
```
anszom: If you mark it as  `total_increasing` , HA will turn it into a meter.

cloud/devices/FAFXU25006.ts
```diff
@@ -0,0 +1,674 @@
...
+                        icon: 'mdi:timer-sand',
+                    },
+                    door: {
+                        platform: 'binary_sensor',
+                        unique_id: '$deviceid-door',
+                        state_topic: '$this/door',
+                        name: 'Door',
+                        device_class: 'door',
+                        payload_on: 'ON',
+                        payload_off: 'OFF',
```
anszom: unnecessary

tests/cloud/devices/FAFXU25006.test.ts
```diff
@@ -0,0 +1,636 @@
...
+)
+
+// Speed Wash: cloud course:"SPEEDWASH", soilWash:"SOILWASH_LIGHT", temp:"TEMP_HOT",
+// spin:"SPIN_EXTRA_HIGH", turboWash:"TURBOWASH_OFF", remainTimeMinute:14.
+const SPEED_WASH = buf(
+    'aaff200a00a2001005000100ec00900a03090e0f2e00000000000000001d001d0000002e0101000200000e2004ffff2d3c200000420004300000000000000400004000000000000000f00064000400000000ffff0100000a010a0e104a00000000000000000e000e0000004a0101000200000e2004ffff2d3c000000420004300000000000000400004000000000000000f00064000400000000ffff0100003e62bb',
+)
+
+// Heavy Duty: cloud course:"HEAVYDUTY", soilWash:"SOILWASH_HEAVY", turboWash:"TURBOWASH_ON",
+// remainTimeMinute:108, a value over 255 minutes' worth of byte, so it exercises the 16-bit read.
```
anszom: 108 < 255. What is this comment about?

It's likely that no program is longer than 255 minutes though.. Injecting manufactured packets to LG could expose how they decode larger values - if that's useful for anything.

anszom (commented):
Thanks! My comments are essentially the same as for #162:

1. Rebase to master & adjust to the changes (`Enum`, 'unknown')
2. Please reduce the comment volume
3. Try to get entity naming & metadata consistent with other devices.

cloud/devices/FAFXU25006.ts
mattst88: Oh, nice. Thanks!

cloud/devices/FAFXU25006.ts
mattst88: Removed, thanks.

tests/cloud/devices/FAFXU25006.test.ts
mattst88: Huh, that is definitely nonsense. Sorry about that. Removed.

mattst88:
Thanks! I've updated the PR (and preemptively my others as well, so you don't have to give the same feedback in multiple places).

anszom:
I've added some adjustments & merged this to master. FYI:
- enum class applied to most entities
- entity names adjusted for consistency with other devices
- pre_state reporting removed - this is not useful in HA

Thanks.

PR #164, opened 2026-09-05
Add support for the LG DLHC5502V heat-pump dryer

mattst88:
Adds a handler for modelId BDH_D30007_US (thinq2 deviceType 202), the
heat-pump dryer sold as DLHC5502V. Read-only.

Frames are the 0xFF-escaped AABB form with 42-byte records, class byte
0x30, against 72 on the matching washer and 28 on the RV13* models,
with the fields at different offsets from either, so this cannot be
aliased. Records carry no marker byte the way the washer's do, so
length and class are the only validation available.

Published: the cycle state and the state before it, the course, dry
level, Eco Hybrid setting and buzzer volume, the remaining time and the
cycle length, the energy the cycle has used, and the Wrinkle Care, Damp
Dry Beep, Detect Load, drum light, condenser clean, remote maintain and
remote start flags.

Confirmed against live capture with rethink bridged to the LG cloud, so
each byte change could be matched against LG's own decoded state at the
same timestamp. Of note:

- The minute counters are big-endian here, the opposite of the matching
  washer's.
- This model's modelJSON enum indices are the wire codes, so the full
  state enum is carried. Nine values were seen live.
- Course has no index table in the modelJSON at all, so all 22 codes
  came from selecting each cycle and reading the byte. Ten are dial
  positions and the other twelve are reachable only from the ThinQ app.
  A 22-entry panelCrsList happens to match that set, but it is not
  proof of completeness: on the matching washer the same field is the
  slots currently assigned to the panel, and it was seen at ten entries
  and then six as they were reassigned.
- The course NAMES are read off the panel rather than taken from LG's
  cloud identifiers, which disagree on most of them and not merely in
  wording. Two cases make the point: the cloud calls code 9 "QUICKDRY"
  while the panel calls it Small Load, and the cycle the panel calls
  Quick Dry is code 23, which the cloud calls "BABYWEAR"; the cloud
  calls code 17 "POWER" while the panel calls it Heavy Duty, Power Dry
  being code 16, which the cloud calls "ALLERGYCARE". The identifiers
  do not merely read awkwardly, they point at the wrong cycles, so
  publishing them would tell an owner their dryer was running something
  it is not. Each entry keeps its identifier as a trailing comment,
  since that is what a capture and the cloud will show.
- A cycle pushed from the app is not a course code at all: rec[5] reads
  the 255 DOWNLOAD sentinel and which cycle it is sits in rec[23], the
  cloud's downloadCourse. Published separately for that reason. One has
  been seen, 114, which the panel calls Wrinkle Prevention.
- rec[5] and rec[20] part company on five cycles, not just the AI
  position, which is what settles rec[20] as the loaded sub-cycle and
  rec[5] as the one to publish.
- The cloud calls condenser cleaning selfCleaning, which is worth
  knowing before hunting for it by name.
- Auto Course Arrange at rec[27] bit 0x01, toggled from the app in both
  directions. It is on by default, which is why that bit reads 1 in
  almost every record captured and had looked like a constant.

rec[26] bit 0x40 is the Remote Start setting, shown by switching it off
at the panel with the machine idle: the byte went 0x44 to 0x04 with
nothing else moving and the cloud reporting REMOTE_START_OFF. What
makes it look ambiguous is that the appliance switches the setting on
by itself: starting a cycle by hand moments later, with the setting
still off, took the byte straight back to 0x44 as the machine entered
DETECTING. The matching LG range names this behaviour in its own
settings, as "Auto Remote": "Starting cooking on the product
automatically turns on Remote Start". This dryer offers no such toggle,
so there is no way to stop it.

Outstanding and unknown:

- rec[16] steps through a small set of values across a run, the same
  sequence on every complete cycle, and also differs per course while
  idle. No cloud field tracks it, so there is no name to publish it
  under.
- rec[27] bits 0x40 and 0x80 are undecoded.
- The door sensor is not there to find: a full 72-field state dump from
  this model carries doorLock but no doorClose, where the matching
  washer's dump carries both.
- The cloud reports error, ductClogging and ductSensingOnOff with no
  wire byte identified. The duct pair would need a blocked vent and the
  error code a real fault.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,7 @@ The following appliances are currently supported in rethink:
     - 🫤 DLE7300WE - preliminary support
     - 👍 DLEX3900B (RV13B6BSD_D_US_WIFI), Electric Dryer - mostly working
     - 👍 RV13B6ES_D_US_WIFI, Electric Dryer - mostly working
+    - 👍 DLHC5502V (BDH_D30007_US), Heat-Pump Dryer - mostly working, read-only
 - WashTowers (combined washer+dryer):
     - 👍 WKEX200HBA (WTL_FXU_BDV_NA_01), WashTower - mostly working
 - Dehumidifiers
--- a/cloud/devices/BDH_D30007_US.ts
+++ b/cloud/devices/BDH_D30007_US.ts
@@ -1,0 +1,465 @@
+import HADevice from './base'
+import { Device as Thinq2Device } from '../thinq2/device'
+import { type Connection } from '../homeassistant'
+import { type Metadata } from '../thinq'
+import { allowExtendedType } from '@/util/casting'
+import AABBDevice from './aabb_device'
+import { Enum } from '@/util/enum'
+
+// LG heat-pump dryer, matched on modelId "BDH_D30007_US" (thinq2 deviceType 202), sold as DLHC5502V.
+//
+// Like its matching washer (see FAFXU25006.ts) this model uses the long, length-escaped AABB framing
+// rather than the short one the RV13* dryers use. A frame on the wire looks like:
+//
+//     aa ff 30 0a 00 66 00 00 7e 00 01 00 ec 00 54 <payload> <chk16> bb
+//      │  │  │  │  └──┬──┘             │  └──┬──┘
+//      │  │  │  │     │                │     └ inner payload length
+//      │  │  │  │     │                └ inner frame type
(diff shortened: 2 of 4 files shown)


anszom:
I've added some adjustments & merged this to master. Generally the same kind of changes as #163. Thanks.

mattst88:
Thank you so much!

PR #165, opened 2026-09-05
Add support for the LG LSIS6338FE electric range

mattst88:
Adds a handler for modelId WLSI_633_ (thinq2 deviceType 301), the
electric range sold as LSIS6338FE. Read-only.

Frames are the 0xFF-escaped AABB form with 115-byte records, class byte
0x40, which no existing handler matches, so this cannot be aliased.

Published: the oven state and cook mode, the target and current
temperatures, the broil level, the oven door and door lock, the control
lock, the kitchen timer, the remaining time and the cycle length, the
five cooktop elements, whether Remote Start is armed and whether the
running cycle was started remotely, and eight appliance settings.

Everything published was confirmed against live capture with rethink
bridged to the LG cloud, so each byte change could be matched against
LG's own decoded state at the same timestamp:

- Nine cook modes, each started at the panel and matched against
  upperManualCookName. Proof needs a cold oven; started warm the
  appliance displays "Hot" and sends nothing.
- The oven state enum, including Cleaning and Cooling from a self clean.
- The five cooktop elements, each lit on its own and named at the panel.
  LG's cloud numbers them in an order that is not the record's: rec[60]
  is cooktop1, rec[63] cooktop4, rec[66] cooktop2, rec[69] cooktop5 and
  rec[72] cooktop3.
- Remote Start, both armed (rec[31] bit 0x01, upperRemoteStart) and in
  use (the top bit of the mode byte, upperRemoteStartUsing). Arming it
  needs the knob turned away from the position and back.
- Eight settings changed one at a time from the app and the panel, each
  a single-byte diff with the cloud naming it in the same frame: the
  clock format, Auto Conversion, the temperature unit, the Temperature
  Adjustment, the Preheating Alarm Light, the oven and cooktop beeper
  volumes, and Auto Remote.

Temperatures need care, because the appliance transmits them in
whichever unit it is set to. The top bit of the target's high byte,
rec[21] 0x80, is a Celsius flag rather than part of the value: read
raw, a 175C bake reports 32943. They are masked and converted to
Fahrenheit before publishing, since the discovery config declares a
fixed unit and cannot follow the appliance. The unit is taken from the
setting byte at rec[2] rather than that flag, because an idle oven has
no target and therefore no flag.

The Temperature Adjustment converts too, but as a difference rather
than an absolute: 1.8 degrees F per degree C with no 32-degree offset.
The app accepts it in either scale and converts to the appliance's unit
before sending, which is what makes it easy to get wrong: with the
appliance in Fahrenheit, entering +4C put 7 on the wire, +10C put 18
and +15C put 27, while in Celsius +4C put a plain 4.

Outstanding and unknown:

- Easy Clean has a knob position but has never been made to start, so
  its code is unknown. Rotating to the position is not enough, the same
  way it was not enough for Remote Start.
- The meat probe has not been exercised at all.
- The cook-mode and oven-state tables hold only values seen live. This
  model's modelJSON enum indices do not match its wire codes, unlike the
  matching washer and dryer, so the rest of each enum cannot be read off
  and would have to be driven.
- Low has never been set on rec[9], the cooktop volume, which offers
  only Mute and High at the panel; that entry carries across from the
  oven's volume, which was watched through all three.
- The second byte of each cooktop block is that element's running time
  in minutes, confirmed on three elements against
  cooktopNOperationTimeMinute. It is not published: it resets to 0 every
  time the element goes out, which makes it useless in HA. The third
  byte of each block has never moved.
- rec[27] bit 0x20 is the cloud's upperTargetTempShow and rec[31] bit
  0x02 its upperTimerSet, the latter matched on seven transitions across
  two captures. Both flip throughout an app conversation, so they are
  documented rather than published.
- rec[11] bit 0x08 tracks the cloud's controlLockEnable, which despite
  the name is not the panel's control lock: it toggles in lockstep with
  three FOTA fields and upperOptionSet whenever the app is talking to
  the appliance. Documented so nobody mistakes it for the lock at bit
  0x01 of the same byte.
- The cloud reports burnerState and burnerOnCounter, a count of lit
  elements, with no wire equivalent found. The clock's time of day is
  not in the record either: syncing it from the app moved nothing.
- Lighting any cooktop element drives the matching LG microwave's hood
  fan to level 1 when that interaction is enabled in the ThinQ app. The
  oven alone never does, not at 425F with the door opened. Nothing of
  this appears in this record; it is noted at the cooktop table because
  it explains a microwave changing state on its own.

The first commit is a separate one-line fix: the device type picker labelled deviceType 301 "Gas Range", which mislabels the electric ranges the type also covers.


anszom:
I've added some adjustments & merged this to master. FYI:
- fixed an inaccurate comment about AABB packet framing
- entity naming consistency corrections

Thanks.

PR #169, opened 2026-09-06
Add WD722RH water purifier support with full captured controls

dohun0310:
Adds support for the LG PuriCare WD722RH water purifier (`1WPD6CMIDR__3`, device type 103) with full control of every setting that could be captured from the ThinQ app, in addition to the existing state decode.

## Controllable settings

All 15 writable entities share the same wire mechanism: a 145-byte sparse config payload (`0xF0 0x17` header, `0xFF` filler, one or two live bytes). The appliance echoes each write into its 270-byte config response in two records; the settled (second) record's offset is always `write offset + 137`, confirmed across all fifteen captured settings rather than assumed from one.

| Home Assistant entity | ThinQ state | Notes |
| --- | --- | --- |
| Ice lock | `wpState.iceLock` | switch |
| Ice-only lever | `wpState.iceLever` | switch |
| Ice-first mode | `wpState.iceFirstModeOnOff` | switch |
| Hot water lock | `wpState.hotWaterLock` | switch, inverted polarity on the wire (LOCK writes 0) |
| Cold water enable | `wpState.coldWaterOnOff` | switch |
| Display brightness | `wpState.LCDBrightness` | select 20/40/60/80/100 (0% not offered on the panel, so not exposed) |
| Time format | `wpState.timeFormat` | select 12/24 |
| Display mode | `wpState.alwaysOnDisplayOnOff` + `clockOnOffInWaitingScreen` | select ALWAYS_ON / WAITING_SCREEN_CLOCK / OFF, written as a mutually-exclusive byte pair |
| Display dimming timeout | `wpState.displayDimmingMin` | number, 1-30 min |
| Display off timeout | `wpState.displayOffMin` | number, 1-30 min |
| Product sound | `wpState.productSoundOnOff` | switch |
| Voice volume | `wpState.voiceVolume` | select 20/40/60/80/100 (0% not offered, same wire literal-percent encoding as brightness) |
| Button sound | `wpState.buttonSoundOnOff` | switch |
| Voice guidance | `wpState.voiceOnOff` | switch |
| Default water type | `wpState.defaultWaterSet` | select RECENT_WATER / NORMAL_WATER / COLD_WATER |
| Default water amount | `wpState.defaultWaterAmountMode` | select over the panel's five dispense presets (120/250/500/1000 mL, last used) |
| Lever dispensing type | `wpState.leverDispensingType` | select PRESSING / CLICK |

Every switch/select/number was captured going both directions (ON→OFF and back, or through every offered option) on the physical unit, and every write test asserts the full generated packet against the real TX capture byte-for-byte.

## Deliberately still read-only

Ice maker, dispensing, and sterilisation remain state-only: an earlier capture proves `iceMaker = OFF` as state, but the ThinQ TX command that produced it was never recorded, so it is not guessed at.

## Tests

- 56 model-specific tests (up from 30)
- new RX fixtures for every newly-decoded setting
- exact full-packet TX assertions for all 15 writable settings
- range/enum rejection tests for uncaptured values (0% brightness/volume, out-of-range minutes, unrecognised select labels)
- `npm run check` passes
- `npm run build` passes
- full suite passes: 500/500

## Scope notes

The README lists the marketing model first: `WD722RH (1WPD6CMIDR__3)`. The handler stays marked preliminary — it is still based on one physical unit — but is no longer read-only.


dohun0310:
Updated the PR with captured write support for three controls on the WD722RH: ice lock, ice-only lever, and ice-first mode.

Each switch now has a real ON/OFF TX pair and a byte-for-byte full-frame test. Model tests pass 30/30; `npm run check`, `npm run build`, and the full 474-test suite pass. Ice-maker control remains deliberately deferred because the earlier recording contains its OFF state but not the command TX frame.

Latest commit: `406d8b4`

dohun0310:
Updated with control for 12 more settings, on top of the ice-related switches from the previous push: hot water lock, cold water enable, display brightness/time-format/mode/timers, product/button sound, voice guidance/volume, default water type/amount preset, and lever dispensing type.

All 15 writable entities go through the same sparse-payload mechanism, and each has a byte-for-byte TX assertion against the real captured frame. Model tests 56/56, full suite 500/500, `npm run check` and `npm run build` pass.

Latest commit: `79cb29d`

dohun0310:
One more setting captured plus an entity-category cleanup:

- **Default hot water temperature** (`wpState.hotWaterIndex` / `defaultHotWaterInfo`) is now a controllable select (40/70/90°C), captured going through all three values and back.
- All the user-facing switches (ice lock, ice-only lever, ice-first mode, hot water lock, cold water enable) now carry `entity_category: 'config'`, matching how `RAC_056905_WW` classifies its jet/energy-saving toggles — previously some were uncategorised (primary) and one was miscategorised as `diagnostic`.
- `child_lock`, a read-only status, is now `diagnostic` for the same reason (previously uncategorised).

Model tests 58/58, full suite 502/502, `npm run check`/`npm run build` pass.

Latest commit: `3bf0c7d`

dohun0310:
Added `dnd_mode` (switch, writable) in 9ee82f4.

**Evidence**: captured live off the panel's do-not-disturb toggle 2026-09-08. Config-frame offset 222 (the settled/second record) flips `0x00 <-> 0x01` on toggle -- everything else in the 266-byte frame is byte-identical between ON and OFF. This is exactly the established `settled = write-offset + 137` convention this driver already uses for every other setting (write offset 85, settled offset 222).

Both directions were captured and are asserted byte-for-byte:
- state decode: `DND_ON`/`DND_OFF` real config frames
- app write: `DND_ON_TX`/`DND_OFF_TX` real TX frames, reproduced via `dev.setProperty('dnd_mode', ...)`

`tsc --noEmit` clean, full suite 504/504 (was 502 before this change).

dohun0310:
Added `dnd_ice_mode` (switch, writable) in 8b61963 -- the sub-toggle inside the do-not-disturb menu that lets the ice maker keep running during DND.

**Evidence**: captured both directions live from the panel. Write offset 127, settled read offset 264 (= 127 + 137), matching the driver's established convention exactly.

One thing worth flagging for reviewers: the app's captured TX frame also carried `dndMode`'s own offset (85) plus two still-unidentified sibling bytes (86-89) as a batch resend of the whole DND submenu -- only offset 127 actually changes between the ON and OFF captures I took. The test fixture isolates just that field via `configCommandRaw`, consistent with how every other single-field toggle in this driver is tested, rather than replaying the noisy multi-field capture verbatim.

`tsc --noEmit` clean, full suite 506/506 (was 504 before this change).

dohun0310:
Added `dnd_start_hour` / `dnd_end_hour` (number, 0-23, writable) in 90d21c4 -- the DND schedule's start/end time pickers.

**Evidence**: captured two independent panel changes (start 23->22, end 06->05). Write offsets 86/88, settled read offsets 223/225 (write + 137), matching the driver's convention exactly.

**Worth a close look**: the panel shows these hours in KST, but the wire carries them in UTC. `raw = (KST hour - 9) mod 24`, verified against both fields across the two captures:
- start 22:00 KST -> wire 13
- end 06:00 KST (before) -> wire 21
- end 05:00 KST (after) -> wire 20

All three match exactly, so I'm confident in the encoding. I deliberately did **not** bake a timezone conversion into the driver -- the appliance's or the user's actual local timezone isn't something rethink can know reliably, and getting that wrong silently would be worse than exposing the raw UTC hour. HA users on non-KST installs will need to account for the offset themselves; happy to add a comment or docs note if that's preferred, or revisit this if there's a better source of truth for the appliance's timezone somewhere else in the protocol.

`tsc --noEmit` clean, full suite 509/509 (was 506 before this change).

dohun0310:
Merged into my local deploy-integration branch alongside the other in-progress device branches; full suite passes there too (710/710), so this is now running on my home Home Assistant instance.

PR #170, opened 2026-09-06
Add support for the FQ18PADNQN floor-standing air conditioner (PAC_910604_WW)

dohun0310:
## Summary

Adds a handler for the LG floor-standing **FQ18PADNQN** (ThinQ model `PAC_910604_WW`, `deviceType` 401), registers it in the Home Assistant bridge, and documents it in the README.

It shares the `0xA7` TLV envelope with the newer RAC units, but the fan encoding, the horizontal-swing field and several feature tags differ enough that it needs its own field map rather than an alias.

## Home Assistant entities

Controllable (every write reproduces the exact command frame the ThinQ app sent, captured owner-operated on the physical unit):

- `climate` - power, `cool` / `dry`, target temperature (18-30 °C), fan (`auto` / `low` / `medium` / `high` / `turbo` / `max`), both swing axes as the climate's own swing modes. A mode/fan/setpoint write repeats the current values in the same frame, exactly as the app does; powering on restores mode, fan and setpoint in one frame, switching off sends the power tag alone, and HA's `off` mode powers the unit down (there is no such wire mode)
- `switch` - air purify, energy saving, smart care (single-tag writes)
- `select` - wind mode (`off` / `coolpower` / `longpower`, mirroring the app's single selector)
- `select` - human sense (`off` / `direct` / `indirect`)
- `number` - sleep timer (minutes, 0-420), turn-on timer `0x21c` and turn-off timer `0x21b` (minutes, 0-1440, 0 cancels)

Reported only:

- power draw in watts (published raw - the RAC `-60` correction does not hold on this model)
- auto-dry enabled state and its remaining countdown; the app's 10/30/60-minute duration setting is cloud-only and produces no local frame
- filter hours used, rated lifetime, computed remaining hours and percentage used; diagnostic error code
- smart guide, PM1.0 / PM2.5 / PM10 and humidity as sensors

Appliance-internal status is filed under `diagnostic`, as on the RAC units; the environmental readings and the controls stay primary. The handler never polls or queries - it is a passive listener that transmits only user-initiated control writes.

## Wind mode findings

The app exposes wind mode as one `off` / `coolpower` / `longpower` selector, and the wire agrees: there is a single shared off frame and no observed `0x236=0` or `0x209=0` write. Coolpower-on is a direct `0x236=1` tag (the unit then drives fan turbo and 18 °C itself); longpower-on is a climate bundle with the fan forced to max (2313, seen twice byte-identical); off is a climate bundle with the fan fixed to high (1542, restoring the previous manual level was tested and does not happen). The implementation sends the bundles with the current mode/target echoed.

## Protocol findings

Every tag was confirmed by operating the physical unit and matching the resulting frames, on-to-off round trips throughout:

| Tag | Meaning | Notes |
|---|---|---|
| `0x1f7` | power | 0 / 1 |
| `0x1f9` | mode | 0 cool, 1 dry |
| `0x1fa` | fan | level duplicated across both bytes (`level × 0x0101`) |
| `0x1fd` / `0x1fe` | current / target temperature | `raw / 2` |
| `0x205` | vertical swing | plain on/off here, not an angle |
| `0x206` | horizontal swing | low byte = right vane, high byte = left |
| `0x208` | human sensing | 0 off, 1 direct, 2 indirect |
| `0x20d` / `0x20f` | energy saving / air purify | single-tag writes |
| `0x20e` / `0x225` | auto-dry enabled / remaining | enabled is 0/1; remaining is minutes; duration selection is not locally observable |
| `0x209` / `0x236` | long power / cool power | reported state; written via the wind-mode selector |
| `0x23e` / `0x23f` | smart care / smart guide | smart care is a single-tag write |
| `0x21a` | sleep timer | minutes, counts down (420 set, 0 cancels) |
| `0x21b` / `0x21c` | turn-off / turn-on reservation | minutes, count down (1440 / 60 set, 0 cancels) |
| `0x2b3` | power draw | watts, 0 while off |
| `0x221` | error code | 0 while healthy |
| `0x355` / `0x356` | filter hours used / rated lifetime | also publishes remaining hours and percent used |
| `0x333` / `0x334` / `0x335` | PM1.0 / PM2.5 / PM10 | |
| `0x336` | humidity | one-to-one percent |

Horizontal swing was established by driving each vane on its own: left alone reports `0x206 = 256`, right alone `1`, off `0`, both `257`. The `both` write echoes that state value - it was never captured as a command. While energy saving engages, one frame carries `0xFF` in one byte of `0x1fa`; the other byte still holds the level, so the fan reading is taken from that byte. `min_temp` / `max_temp` are 18-30 (18 °C forced by coolpower is the observed floor; above the observed 26 °C ceiling the range is unverified). The `0x25e = 2` accompanying smartcare is still not exposed - its precise meaning is unconfirmed.

## Compared with the other AC PRs

Checked against #36, #60, #73 and #129 for anything missed. PAC now exposes every RAC-style feature that has an exact PAC tag or captured behavior: climate controls, both swing axes, air purify, energy saving, timers, auto-dry enabled/countdown, error, power draw and filter counters. RAC-only capability-gated diagnostics (EEV, pipe/ODU temperatures, fan RPM and nominal capacity) are not added because none of their RAC tags has been identified on this PAC capture. The RAC filter reset button is also deliberately excluded: RAC uses a private query/reset protocol, while PAC reports `0x355`/`0x356` in ordinary state frames and no PAC reset command was captured. Writing zero to a counter or copying RAC's private reset would be an unsafe guess.

Two differences worth flagging: #129 divides `0x336` by 10 for humidity on a cassette unit, but on this model the same tag is a one-to-one percentage and matches both the app and a reference hygrometer; and the reservations here count down in minutes on the wire (`0x21b`/`0x21c`), so the number entities use minutes rather than the RAC hour sliders.

## Tests

Fixtures are real captured frames from two owner-operated sessions; the suite validates their declared envelope lengths and CRCs, then covers immediate discovery and delta-first reconnects, state decoding (power, all modes and fan levels including the energy-saving override, both swing axes on the climate entity, human sensing, feature toggles, auto-dry enabled state, all three timers, power draw, raw/rated/remaining filter life and percentage recomputation with a zero rated lifetime and clamping, delta-only tags, malformed input), that every control write reproduces the app's captured frame (climate mode/fan/setpoint/power/attach rules, both swing axes, switches, human sense, all three timers including cancel, wind-mode coolpower/longpower/off), that sensor writes still send nothing, and the diagnostic/primary split.

Verified locally: `npm run check`, `npm run build`, `npm test` (496/496).

### DIFF
--- a/cloud/devices/PAC_910604_WW.ts
+++ b/cloud/devices/PAC_910604_WW.ts
@@ -1,0 +1,359 @@
+import TLVDevice from './tlv_device'
+import { Device as Thinq2Device } from '../thinq2/device'
+import { ClimateComponent, DeviceDiscovery, type Connection } from '../homeassistant'
+import { type Metadata } from '../thinq'
+import { allowExtendedType } from '@/util/casting'
+import { Enum } from '@/util/enum'
+import * as TLV from '@/util/tlv'
+import HADevice from './base'
+
+/**
+ * LG floor-standing air conditioner PAC_910604_WW (deviceType 401).
+ *
+ * Same TLV envelope as the RAC_056905_WW wall units (0xA7 header), but the field
+ * set is incompatible, so it gets its own handler instead of an alias:
+ * fan levels replicate across both bytes (level x 0x0101), horizontal swing is a
+ * two-bit field, and human-sense / power modes / air-quality sensors are new tags.
+ *
+ * Read-only: the handler never transmits. All discovery entries are state-only
+ * and setProperty() on any of them is a tested no-op.
+ */
+
+// Owner-verified on the physical unit (each toggle driven on the remote/app and
+// confirmed on the wire, on->off round trips; PM/humidity cross-checked with the
+// ThinQ app): 0x1f7 power, 0x1f9 mode (cool/dry only), 0x1fd/0x1fe temperatures
+// (/2), 0x205 vertical swing, 0x206 horizontal swing, 0x208 human sense,
+// 0x20d eco, 0x20f air clean, 0x21a sleep timer (minutes, up to 420),
+// 0x225 auto-dry countdown (minutes), 0x236 cool power, 0x209 long power,
+// 0x23e smart care, 0x23f smart guide,
+// 0x333/0x334/0x335 PM1.0/PM2.5/PM10, 0x336 humidity.
+const MODES = Enum.of({
+    cool: 0,
+    dry: 1,
+})
+
+// Observed levels: 2 (low), 4 (medium), 6 (high), 8 (auto),
+// 7 (coolpower only), 9 (longpower only).
+const FAN_LEVELS = new Enum([
+    ['auto', 8],
+    ['low', 2],
+    ['medium', 4],
+    ['high', 6],
+    ['turbo', 7],
+    ['max', 9],
+] as [string, number][])
+
+// Bit-split field: low byte = right vane, high byte = left vane.
+const SWING_H = Enum.of({
+    off: 0,
+    right: 1,
+    left: 256,
+    both: 257,
+})
+
+const HUMAN_SENSE = Enum.of({
+    off: 0,
+    direct: 1,
+    indirect: 2,
+})
+
+/**
+ * Fan values arrive as level x 0x0101 (both bytes equal: 0x0202, 0x0404 ...).
+ * While eco is engaging the device emits one frame with 0xFF in one of the
+ * bytes (0xFF04 or 0x04FF observed); the other byte still carries the level.
+ */
+function fanLevel(raw: number): number {
+    const lo = raw & 0xff
+    const hi = (raw >> 8) & 0xff
+    return lo === 0xff ? hi : lo
+}
+
+export default class Device extends TLVDevice {
+    meta: Metadata
+    configDone = false
+
+    constructor(HA: Connection, thinq: Thinq2Device, meta: Metadata) {
+        super(HA, thinq)
+        this.meta = meta
+        // Stay fully passive: never poll the appliance, only listen.
+        if (this.query_caps_timeout != undefined) {
+            clearInterval(this.query_caps_timeout)
+            this.query_caps_timeout = undefined
+        }
+    }
+
+    queryCaps() {
+        // Read-only: no capability query is ever sent.
+    }
+
+    query() {
+        // Read-only: no values query is ever sent.
+    }
+
+    start() {
+        // Read-only: no periodic refresh timer.
+    }
+
+    isValuesResponse(tlvArray: TLV.TLV[]) {
+        return tlvArray.some(({ t }) => t === 0x1f7)
+    }
+
+    valuesReceived() {
+        if (this.configDone) return
+        this.configDone = true
+        this.initConfig()
+    }
+
+    getPowerTLV() {
+        return this.raw_clip_state[0x1f7]
+    }
+
+    getModeTLV() {
+        return this.raw_clip_state[0x1f9]
+    }
+
+    initConfig() {
+        const config: DeviceDiscovery & { components: { climate: ClimateComponent } } = allowExtendedType({
+            ...HADevice.config(this.meta, { name: 'LG Stand Air Conditioner' }),
+            components: {
+                climate: {
+                    platform: 'climate',
+                    unique_id: '$deviceid-climate',
+                    name: null,
+                    temperature_unit: 'C',
+                    // Setpoints move in whole degrees (raw steps of 2).
+                    temp_step: 1,
+                    precision: 1,
+                    current_humidity_topic: '$this/humidity-',
+                    // 18C forced by coolpower is the observed floor; above the
+                    // observed 26C ceiling the range is unverified.
+                    min_temp: 18,
+                    max_temp: 30,
+                    modes: ['off', ...MODES.options],
+                    fan_modes: FAN_LEVELS.options,
+                } satisfies ClimateComponent,
+            },
+        })
+
+        this.addField(config, {
+            id: 0x1fd,
+            name: 'current_temperature',
+            comp: 'climate',
+            state_topic: 'topic',
+            writable: false,
+            read_xform: (raw) => raw / 2,
+        })
+        this.addField(config, {
+            id: 0x1fe,
+            name: 'temperature',
+            comp: 'climate',
+            writable: false,
+            read_xform: (raw) => raw / 2,
+        })
+        this.addField(config, {
+            id: 0x1f7,
+            name: 'power',
+            comp: 'climate',
+            readable: false,
+            writable: false,
+            read_xform: (raw) => (raw ? 'ON' : 'OFF'),
+            read_callback: (val) => {
+                this.HA.publishProperty(this.id, 'power-', val)
+                // Re-evaluate 'mode' so it flips between 'off' and the set mode.
+                this.processKeyValue(0x1f9, this.raw_clip_state[0x1f9])
+                return false
+            },
+        })
+        this.addField(config, {
+            id: 0x1f9,
+            name: 'mode',
+            comp: 'climate',
+            writable: false,
+            read_xform: (raw) => {
+                if (this.getPowerTLV() === 0) return 'off'
+                return MODES.map(raw)
+            },
+        })
+        this.addField(config, {
+            id: 0x1fa,
+            name: 'fan_mode',
+            comp: 'climate',
+            writable: false,
+            read_xform: (raw) => FAN_LEVELS.map(fanLevel(raw)),
+        })
+
+        // Binary sensor fields
+        const powerComp = {
+            platform: 'binary_sensor',
+            unique_id: '$deviceid-power',
+            name: 'Power',
+            state_topic: '$this/power-',
+        }
+        config['components']['power'] = powerComp
+
+        const binaryFields = [
+            { id: 0x205, name: 'swing_v', desc: 'Vertical swing' },
+            { id: 0x20d, name: 'eco', desc: 'Energy saving' },
+            { id: 0x20f, name: 'airclean', desc: 'Air purify' },
+            { id: 0x23f, name: 'smartguide', desc: 'Smart guide' },
+            { id: 0x236, name: 'coolpower', desc: 'Cool power' },
+            { id: 0x209, name: 'longpower', desc: 'Long power' },
+            { id: 0x23e, name: 'smartcare', desc: 'Smart care' },
+        ]
+        for (const f of binaryFields) {
+            config['components'][f.name] = {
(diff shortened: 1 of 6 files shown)


dohun0310:
Added since the initial submission, both confirmed on the physical unit:

- **Auto dry enable** (`0x20e`) — the owner toggled the setting off and back on in the app; only this tag moves. The 10/30/60 minute duration the app offers is never reported on the wire, so only the on/off state is exposed.
- **Filter usage** (`0x355` hours in service / `0x356` rated lifetime, 720 h) — published as a used percentage. The app shows the same pair as "10 h used / 710 h remaining" and renders 2%, matching the computed ratio. A zero rated lifetime is guarded and the value is clamped to 0–100.
- **Error code** (`0x221`) — exposed as a diagnostic sensor, 0 while healthy.

Both new sensors are `entity_category: 'diagnostic'` and everything remains read-only. Local verification: `npm run check`, `npm run build`, `npm test` (485 passing).

dohun0310:
Force-pushed a rework: this PR is now a **read-only** handler. The previous revision had a write path, but several writes had no captured command frame to verify against, so everything is state-only now (no queries, no refresh timer, writes are a tested no-op). Power draw / error code / filter usage from the old revision are dropped in this version. PR body updated to match. Verified: `npm run check`, `npm run build`, `npm test` (480/480).

dohun0310:
Pushed the control implementation. Air purify / energy saving / smart care are switches, wind mode and human sense are selectors, and sleep + both reservations are minute numbers - every write reproduces the app's captured frame byte-for-byte (tests assert this). Notable wire finding: wind mode is a single off/coolpower/longpower selector in the app, with one shared off frame (fan fixed to high) and no standalone 0x236=0/0x209=0 write. Climate, swings and sensors stay state-only. Verified: `npm run check`, `npm run build`, `npm test` (487/487).

dohun0310:
Follow-up: full climate control is restored on the climate entity (power/mode/fan/setpoint with the app's attach rules, both swing axes as climate modes), plus power draw, filter-used and error sensors. Reason for the round trip: the intermediate read-only revision was a caution step while the remaining controls were being captured; all writes here reproduce captured app frames, asserted byte-for-byte in the tests. One documented approximation: the horizontal 'both' (257) write echoes the observed state value, it was never captured as a command. Verified: `npm run check`, `npm run build`, `npm test` (493/493).

dohun0310:
Added the requested PAC follow-up in `518df77`: restored owner-captured auto-dry enabled state, added filter used/rated/remaining hours alongside percent used, and made discovery immediate so delta-first reconnects populate HA. I also audited the RAC handler feature-by-feature. I did not expose the RAC filter-reset button: RAC resets through a private query/reset protocol, while this PAC reports ordinary `0x355`/`0x356` state tags and no PAC reset command has been captured; copying the RAC command or writing zero would be an unsafe guess. Likewise EEV, pipe/ODU temperature, fan-RPM and nominal-capacity sensors remain excluded because their RAC tags are not identified in PAC captures. Verified: `npm run check`, `npm run build`, `npm test` (496/496).

dohun0310:
Small follow-up: aligned entity_category with RAC_056905_WW's convention. Air purify, energy saving, smart care, wind mode and human sense now carry `entity_category: 'config'` (they previously had none, which put them alongside `climate` as if they were the AC's primary control) — matching how RAC classifies its jet/energysave/airclean toggles. Timers stay primary, matching RAC's own timer numbers; diagnostics are unchanged.

Model tests 30/30, full suite 496/496, `npm run check`/`npm run build` pass.

Latest commit: `09f3b96`

cloud/devices/PAC_910604_WW.ts
```diff
@@ -0,0 +1,689 @@
...
+        const publishFilterUsed = () => {
+            const used = this.raw_clip_state[0x355]
+            const max = this.raw_clip_state[0x356]
+            if (used != null && max != null && max > 0) {
+                const remaining = Math.max(0, max - used)
+                const percent = Math.max(0, Math.min(100, Math.round((used / max) * 100)))
+                this.HA.publishProperty(this.id, 'filter_used_time-', used)
+                this.HA.publishProperty(this.id, 'filter_life_time-', max)
+                this.HA.publishProperty(this.id, 'filter_remaining-', remaining)
+                this.HA.publishProperty(this.id, 'filter_used-', percent)
```
maciejsszmigiero: I think it makes no sense to provide both raw values (filer used time and life time) AND calculated values of filter remaining and filter used percentage.

If someone truly wants these values they can simply create a Home Assistant template sensor calculating these and attach it to the AC HA device.

cloud/devices/PAC_910604_WW.ts
```diff
@@ -0,0 +1,689 @@
...
+    query() {
+        // State arrives unprompted, so no values query is sent.
+    }
+
+    start() {
+        // Read-only: no periodic refresh timer.
+    }
+
+    isValuesResponse(tlvArray: TLV.TLV[]) {
+        return tlvArray.some(({ t }) => t === 0x1f7)
```
maciejsszmigiero: Missing length check, could accidentally treat power switch notification as values response.

See `isValuesResponse()` in `RAC_056905_WW.ts` how this should be done instead.

maciejsszmigiero (changes requested):

cloud/devices/PAC_910604_WW.ts
```diff
@@ -0,0 +1,689 @@
...
+        // reconnect that receives only delta frames populate Home Assistant.
+        this.valuesReceived()
+    }
+
+    queryCaps() {
+        // The exact-model field map is fixed; capabilities are not queried.
+    }
+
+    query() {
+        // State arrives unprompted, so no values query is sent.
```
maciejsszmigiero, line 123: Are you sure that this device truly sends its values response on its own and not in a reply to a query from the ThinQ cloud connected via bridge?

cloud/devices/PAC_910604_WW.ts
```diff
@@ -0,0 +1,689 @@
...
+            this.query_caps_timeout = undefined
+        }
+        // The field map is fixed for this exact model, so discovery does not
+        // need to wait for a full frame containing power. This also lets a
+        // reconnect that receives only delta frames populate Home Assistant.
+        this.valuesReceived()
+    }
+
+    queryCaps() {
+        // The exact-model field map is fixed; capabilities are not queried.
```
maciejsszmigiero, line 119: Does the device respond in some way to the usual 0x1f5/1 caps query?

If yes, then I would keep the normal caps query flow rather than hack around it in the constructor.

dohun0310:
Addressed review feedback:

- **Filter (time only per owner request):** Removed the calculated `filter_used` percentage — kept the three duration sensors `filter_used_time` / `filter_life_time` / `filter_remaining` (h, diagnostic). Users can derive % via a template sensor if needed. Updated `PAC_910604_WW.ts` and its tests.
- **`isValuesResponse` length guard:** Now `tlvArray.length >= 10 && tlvArray.some(({t}) => t === 0x1f7)` mirroring `RAC_056905_WW`, so a single-tag power notification is not mistaken for a values response.
- **Passive listener vs caps query (questions on lines 106/110):** Verified against bridge `clip/message` captures: state frames (A7 envelope) arrive unprompted without a prior `0x1f5` values query, and a caps query `0x1f5 v1` gets no response — field map is fixed for this exact model, so the handler suppresses polling and publishes discovery on the first values frame. No periodic refresh is needed.

Pushed to `stand-ac-PAC_910604_WW` (036fea3). `npm run check && npm run build && npm test` — 496/496.

cloud/devices/PAC_910604_WW.ts
dohun0310: Agreed. I removed the calculated percentage entity and kept only the hour-based filter sensors (`filter_used_time`, `filter_life_time`, and `filter_remaining`). The tests now assert that `filter_used` is not published. Addressed in 036fea3.

cloud/devices/PAC_910604_WW.ts
dohun0310: Fixed in 036fea3. `isValuesResponse()` now requires `tlvArray.length >= 10` in addition to the power tag, matching `RAC_056905_WW`, so a power-only notification cannot be mistaken for a values response.

cloud/devices/PAC_910604_WW.ts
dohun0310: Yes. I verified this from the bridge-side `clip/message` capture: the PAC sends A7 state frames without a preceding values query from rethink. Both full frames and changed-tag notifications continue arriving passively, including while the handler sends no periodic queries.

cloud/devices/PAC_910604_WW.ts
dohun0310: I tested the usual `0x1f5=1` capabilities query and did not observe a capabilities response from this exact model. Since its fixed model-specific field map is established from captured frames and state arrives unprompted, I kept the passive constructor path rather than the normal caps-query flow.

dohun0310:
Added `air_quality` sensor (7109e55), the overall indoor air quality grade the panel shows alongside the PM 1.0/2.5/10 readings.

**Evidence**: tag 0x240 read 1 at capture start; the owner confirmed the panel showed "Good" at that point. Then spraying an aerosol in front of the unit made the panel jump straight to "Very bad", and the tag jumped to 4 in the same wire frame, with no other value seen for it anywhere else in the session.

**Caveat worth flagging**: only the two endpoints (1 and 4) were directly observed. The owner confirmed the panel uses a 4-grade scale (good/moderate/bad/very bad), so 2 and 3 are inferred from that ordering rather than independently captured -- I didn't have an easy way to land the unit exactly on the middle grades. If a later capture catches 2 or 3 on the wire and they don't match this ordering, the enum mapping in `AIR_QUALITY` is the only thing that needs to change.

State-only (writable: false) since no command frame for this tag was ever seen -- the app has no control for it.

**On CO2/odor** (also tracked in the gaps list): investigated, but the owner confirmed this unit's panel air-quality screen only shows the 3 PM readings plus this overall grade -- no CO2 or odor display exists here, even though the modelJSON has `airState.quality.CO2`/`CO2Level`/`odor` fields (same superset-noise situation as the PAC child-lock field from an earlier PR). Not adding anything for those, since there's no UI reference to confirm the wire ever carries them for this SKU.

`tsc --noEmit` clean, full suite 497/497 (was 496 before this change).

dohun0310:
Merged into my local deploy-integration branch alongside the other in-progress device branches; full suite passes there too (710/710), so this is now running on my home Home Assistant instance.

anszom:
FYI, I'm deferring all AC work until a common base gets merged either in the form proposed in #122 or another. This may take some time though, I'm a bit overloaded right now.

PR #171, opened 2026-09-07
Add LG WM4000HWA (F3P2CYUBE__) front-load washer

rtvanhook:
Adds the WM4000HWA (modelId `F3P2CYUBE__`, ThinQ2 / deviceType 201) as a self-contained device module, in the same one-file-per-model shape as the rest of `cloud/devices/`.

**New files**
- `cloud/devices/F3P2CYUBE__.ts` — decoder + control
- `tests/cloud/devices/F3P2CYUBE__.test.ts` — 23 tests; every fixture is a real captured frame
- one line in `README.md`

**What it covers**
- **Read** — the full status record: power/state, course (12 dial positions + Downloaded slot), soil/temp/spin/rinse, remaining & cycle time, delay-wash, cold/turbo/pre-wash/steam/Rinse+Spin, FreshCare, Signal, remote-start & child-lock, Add Item, AI DD, cycles-since-clean, energy — 26 fields.
- **Control** — pause, resume, power-off (WMOff, as a guarded select), and `start_course` (config/start for any of the 13 courses at its defaults, with a 0–19h delay-start folded in). Every write is an exact packet captured from the LG cloud in bridge mode, re-checksummed by `send()`.
- **SmartCourse downloads** — all 17 specialty cycles (Baby Clothes, Denim, EconoWash, Overnight, …) as a `specialty` select that fires each course's captured WMDownload frame into the `0xFF` slot, plus a `downloaded_course` sensor that reads back which is loaded (rec[24]).

**How it was built** — reverse-engineered with `rethink-capture --cloud`, using the LG cloud's own decoded values as a time-aligned oracle against the raw AABB frames; each offset/bit was pinned by toggling one variable on the panel and watching the byte move. The 17 download frames are the exact bytes captured while pushing each SmartCourse from the app — a test asserts all 17 re-frame byte-identical.

**Deliberately out of scope** (grammar known, documented in-file): the full per-field remote cycle builder (course-constrained option availability makes it combinatorial) and power-on (WMWakeup). The physical door lock isn't in the protocol — LG doesn't expose it, and its own app shows no lock tile.

Same AABB frame family as the WM3900HBA (`F3L2CYU__`) sibling, but a longer 44-byte record, so the offsets are its own.

https://claude.ai/code/session_01RDBjWcdSEcgCB2LmjcD8xr


rtvanhook:
Documentation for this device is up on my fork's wiki, since the project wiki isn't writable from outside:

- https://github.com/rtvanhook/rethink/wiki/Appliance:F3P2CYUBE__
- https://github.com/rtvanhook/rethink/wiki/Developing-a-device-against-the-stock-image

The first is the protocol writeup in the `Appliance:<modelId>` convention, linking `AABBProtocol` for framing: the 44-byte status record, enums, the inbound command grammar with captured packets, and the `WMDownload` channel with all 17 SmartCourse codes and frames. It also records the negative results — notably that this model has no usable door or door-lock signal, unlike the `F3L2CYU__` sibling — so nobody repeats that hunt.

The second is a general page, not device-specific: how to develop a device against the stock `ghcr.io/anszom/rethink:dev` image by bind-mounting the built device file and `ha_bridge.js` over `/app/dist`, rather than rebuilding an image each iteration. The wiki covers writing a device and installing rethink, but nothing in between, and this is the part I had to work out to test against my own machine.

Copy either into the project wiki as-is if they're useful.

Also note this PR now has a second commit restoring a bare start verb (`f0e5000201ff010301`) that I'd wrongly removed — it runs whatever cycle is set at the panel, which sidesteps the per-course option-validity problem entirely.

rtvanhook:
Cross-reference: #145 is the same model ID (WM4000HBA, the black variant). Its 0xDE frame is this same
  44-byte record behind the 3-byte header, captured before a startup request. I've asked @nickdavies there
  to run this handler against his unit.

rtvanhook:
Rebased onto current master (README conflict from the device-list rearrangement resolved as one row in the new Washing machines table) and added the missing registration of `F3P2CYUBE__` in `ha_bridge.ts` — the handler had never been added to the model map, so the bridge could not dispatch the washer to it. Device tests pass (23/23). #145 is the same model ID; this handler covers the full 44-byte record and the course downloads, so it should supersede it if you take this one.

PR #172, opened 2026-09-07
Add support for the AS309DWA air purifier (AIR_910604_WW)

dohun0310:
## Summary

Adds support for the LG PuriCare **AS309DWA** air purifier (ThinQ model `AIR_910604_WW`, `deviceType` 402), registers it in the Home Assistant bridge, and documents it in the README.

The appliance uses the `0xA7` TLV protocol and sends both full state reports and short changed-tag notifications.

## Home Assistant entities

Controllable:

- `fan` — power and main fan speed (low / medium / high / turbo / auto)
- `select` — operating mode: Clean Booster, single clean, dual clean, auto
- `select` — Clean Booster circulation fan speed
- `switch` — circulation fan rotation, clean indicator light, air sterilization
- `number` — sleep timer in minutes

Reported only:

- appliance air-quality and odour indices
- PM1.0 / PM2.5 / PM10
- remaining time and rated lifetime for both filters, reported directly in hours
- diagnostic error code

Temperature (`0x1fd`) and humidity (`0x336`) are deliberately omitted because prior exact-model testing found that neither tracks room conditions.

## Protocol findings

Each writable tag was confirmed by operating the physical unit and matching the resulting command frame. Tests assert that writes reproduce the captured TLVs.

| Tag | Meaning | Values |
|---|---|---|
| `0x1f7` | power | 0 / 1 |
| `0x1f9` | mode | 13 Clean Booster, 14 single clean, 15 dual clean, 16 auto |
| `0x1fa` | main fan | 2 low, 4 medium, 6 high, 7 turbo, 8 auto |
| `0x326` | Clean Booster fan | same levels |
| `0x327` | circulation rotation | 0 / 1 |
| `0x24e` | clean indicator | 0 / 1 |
| `0x360` | air sterilization | 0 / 1 |
| `0x21a` | sleep timer | minutes, counts down |
| `0x240` / `0x241` | air-quality / odour index | read-only |
| `0x333` / `0x334` / `0x335` | PM1.0 / PM2.5 / PM10 | read-only |
| `0x355` / `0x356` | lower filter remaining / rated lifetime | hours |
| `0x363` / `0x364` | top filter remaining / rated lifetime | hours |
| `0x221` | error code | read-only |

The product UI names mode 13 “Clean Booster” and mode 14 “Single clean”, so those names are used instead of the generic model-metadata labels.

## Appliance quirks

1. A write received while the unit is off is acknowledged and ignored. Mode and fan writes therefore attach the power tag and start the unit in the requested state. Switching off attaches nothing.
2. In auto mode the appliance controls the fan itself and ignores a bare fan write. The handler does not silently change the operating mode.
3. Setting the sleep timer switches the clean indicator off. This is appliance behavior and is only reported by the handler.

## Passive state handling

Bridge-side captures confirm that the purifier sends A7 state frames and changed-tag notifications without a preceding values query. The normal `0x1f5=1` capabilities query produced no response for this exact model, so the handler uses its captured, fixed field map and does not poll.

`isValuesResponse()` includes the same minimum-TLV-count guard used by `RAC_056905_WW`, preventing a power-only notification from being classified as a full values response.

## Prior exact-model work

The field-name baseline, main-fan captures, filter semantics, and appliance quirks were cross-checked against:

- https://github.com/sosohage2/rethink-korea/commit/c3cf55975c913e206586b7e883b88fe12687a70c
- https://github.com/sosohage2/rethink-korea/commit/387be48a1be28b02dd9691daf58b903a78435287

This PR additionally contributes newly captured Clean Booster fan (`0x326`) and circulation rotation (`0x327`) controls, product-UI mode names, and delta-first discovery.

## Tests

Real captured fixtures are validated for envelope length and CRC. The suite covers:

- all four modes and independent main/Clean Booster fan speeds
- power, rotation, indicator, sterilization, and sleep timer state/write paths
- raw filter remaining/rated lifetime values in hours for both filters
- delta-first discovery and preservation of omitted state
- malformed frames and read-only sensor behavior
- bridge routing for the exact model ID
- `isValuesResponse()` rejection of short power notifications

Verified locally: `npm run check`, `npm run build`, `npm test` (**481/481**).

### DIFF
--- a/cloud/devices/AIR_910604_WW.ts
+++ b/cloud/devices/AIR_910604_WW.ts
@@ -98,7 +98,7 @@ export default class Device extends TLVDevice {
     }
 
     isValuesResponse(tlvArray: TLV.TLV[]) {
-        return tlvArray.some(({ t }) => t === TAG_POWER)
+        return tlvArray.length >= 10 && tlvArray.some(({ t }) => t === TAG_POWER)
     }
 
     valuesReceived() {
--- a/cloud/devices/AIR_910604_WW.ts
+++ b/cloud/devices/AIR_910604_WW.ts
@@ -1,0 +1,339 @@
+import TLVDevice from './tlv_device'
+import { Device as Thinq2Device } from '../thinq2/device'
+import { DeviceDiscovery, type Connection } from '../homeassistant'
+import { type Metadata } from '../thinq'
+import { allowExtendedType } from '@/util/casting'
+import { Enum } from '@/util/enum'
+import * as TLV from '@/util/tlv'
+import HADevice from './base'
+
+/**
+ * LG PuriCare air purifier AIR_910604_WW (deviceType 402).
+ *
+ * The tag-name baseline, filter semantics and the write quirks below were
+ * cross-checked against the earlier exact-model work in sosohage2/rethink-korea
+ * commits c3cf559 and 387be48. That work's correction not to publish the
+ * appliance's internal temperature (0x1fd) and humidity (0x336) as room
+ * measurements is retained: neither tracks the room.
+ *
+ * Every writable field here was driven from the app and the resulting command
+ * frame captured, so the values written are the ones LG itself sends:
+ *
+ *   0x1f7  power                 0/1
+ *   0x1f9  operating mode        13 clean booster, 14 single, 15 dual, 16 auto
+ *   0x1fa  main fan speed        2 low, 4 medium, 6 high, 7 turbo, 8 auto
+ *   0x326  clean booster fan     same levels
+ *   0x327  circulation rotation  0/1
+ *   0x24e  clean indicator       0/1
+ *   0x360  air sterilization     0/1
+ *   0x21a  sleep timer           minutes, counts down
+ *
+ * Two appliance quirks, both observed rather than worked around:
+ *
+ * 1. A write that arrives while the unit is off is acknowledged and then
+ *    ignored. Sending the power tag alongside does work and starts the unit
+ *    directly in the requested state, so mode and fan writes force power on and
+ *    attach it.
+ * 2. While the operating mode is auto the appliance drives the fan itself and a
+ *    bare fan write is ignored. Nothing is done about this - silently changing
+ *    the user's mode to make a fan request stick would be worse.
+ *
+ * Setting the sleep timer also switches the clean indicator off; the appliance
+ * reports both tags in one frame. That is its own behaviour, not something done
+ * here.
+ */
+const MODES = new Enum([
+    ['clean_booster', 13],
+    ['single_clean', 14],
+    ['dual_clean', 15],
+    ['auto', 16],
+] as [string, number][])
+
+const FAN_SPEEDS = new Enum([
+    ['low', 2],
+    ['medium', 4],
+    ['high', 6],
+    ['turbo', 7],
+    ['auto', 8],
+] as [string, number][])
+
+const TAG_POWER = 0x1f7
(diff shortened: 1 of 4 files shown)


cloud/devices/AIR_910604_WW.ts
```diff
@@ -0,0 +1,339 @@
...
+    query() {
+        // State arrives unprompted, so no values query is sent.
+    }
+
+    start() {
+        // No periodic refresh timer.
+    }
+
+    isValuesResponse(tlvArray: TLV.TLV[]) {
+        return tlvArray.some(({ t }) => t === TAG_POWER)
```
maciejsszmigiero: Same remark as for `isValuesResponse()` in #170.

cloud/devices/AIR_910604_WW.ts
```diff
@@ -0,0 +1,339 @@
...
+        // frames is still usable.
+        this.valuesReceived()
+    }
+
+    queryCaps() {
+        // Capabilities are not queried; the exact-model field map is fixed.
+    }
+
+    query() {
+        // State arrives unprompted, so no values query is sent.
```
maciejsszmigiero, line 93: Same remark as to #170.

cloud/devices/AIR_910604_WW.ts
```diff
@@ -0,0 +1,339 @@
...
+            clearInterval(this.query_caps_timeout)
+            this.query_caps_timeout = undefined
+        }
+        // Publish discovery immediately so a connection that only sees delta
+        // frames is still usable.
+        this.valuesReceived()
+    }
+
+    queryCaps() {
+        // Capabilities are not queried; the exact-model field map is fixed.
```
maciejsszmigiero, line 89: Same remark as to #170.

maciejsszmigiero (changes requested):

dohun0310:
Addressed review feedback (same as #170):

- **`isValuesResponse` length guard:** Now `tlvArray.length >= 10 && tlvArray.some(({t}) => t === 0x1f7)`, matching `RAC_056905_WW` — avoids treating a single-tag power notification as a full values response.
- **Passive listener / caps query (lines 89/93):** State frames arrive unprompted via the bridge (A7 envelope) without a prior `0x1f5` values query, and `0x1f5 v1` caps query gets no response for this exact model. Handler keeps a fixed field map and suppresses polling, publishing discovery on the first values frame (no periodic refresh).

Pushed to `air-purifier-AIR_910604_WW` (a019d68). `npm run check && npm run build && npm test` pass.

cloud/devices/AIR_910604_WW.ts
dohun0310: Fixed in a019d68. `isValuesResponse()` now requires `tlvArray.length >= 10` as well as the power tag, matching `RAC_056905_WW`, so power-only notifications are not treated as full values responses.

cloud/devices/AIR_910604_WW.ts
dohun0310: Confirmed from the bridge-side `clip/message` capture: this purifier sends A7 state frames and changed-tag notifications without a preceding values query from rethink. State continues arriving with the handler operating passively.

cloud/devices/AIR_910604_WW.ts
dohun0310: I tested the normal `0x1f5=1` capabilities query and did not observe a capabilities response from this exact model. Its field map is fixed from captured frames and values arrive unprompted, so I kept the passive path rather than the normal caps-query flow.

PR #173, opened 2026-09-07
Add LG S5MPC Styler support

dohun0310:
## Summary

Adds first-class support for the **LG S5MPC Styler** (`deviceType: 203`). S5MPC uses AA/BB framing but has its own 27-byte state record, so it is implemented as a dedicated handler rather than an alias of the existing 40-byte Styler parser.

The implementation is based on traffic captured from the physical appliance. It supports native courses, all model-declared downloadable courses, completion reservations, Store mode, captured remote controls, Home Assistant state reporting, restart-safe smart-course tracking, completed countdown correction, and this-cycle energy.

## What is supported

- 19 selectable native courses
- All 26 downloadable smart courses declared by S5MPC
- Immediate and reserved course starts (0 or 3 to 19 hours; 1-2h rejected per LG's declared range)
- Downloadable-course install and start using the same two-frame sequence sent by LG's app
- Pause, resume, remote power off, and Store mode
- Power, status, course, smart course, duration, remote start, child lock, errors, diagnosis, and energy sensors

No command is included only because another Styler supports it. Every transmitted operation has a matching S5MPC capture or is constructed from this model's own `ControlConvertingRule` and validated against captured command frames.

## State protocol

State frames contain one or more length-prefixed 27-byte records. The final record is current; earlier records are historical and are not published.

Confirmed fields include:

- `record[0]`: top-level state
- `record[1..2]`: remaining hour/minute
- `record[3..4]`: initial hour/minute
- `record[5]`: native or base course
- `record[6]`: error
- `record[12..13]`: reservation time
- `record[14]` bits: remote start and physical child lock
- `record[17..18]`: this-cycle energy, 16-bit big-endian Wh
- `record[20]`: downloadable smart-course ID

Malformed frames, short notifications, ACKs, and downloadable-course name lists are ignored.

## Energy

`record[17..18]` is this-cycle cumulative energy in Wh, not instantaneous power. During a full 104-minute steam cycle the field remained zero while running, changed to `0x0251` at completion, and remained 593 in the following `31e2` and `31eb` frames. The ThinQ app showed exactly 593Wh for the same day.

Home Assistant discovery now declares:

- `device_class: energy`
- `unit_of_measurement: Wh`
- `state_class: total_increasing`

## Remaining time at completion

The physical Styler keeps reporting `00:01` after the cycle has completed and while powered off. The handler now publishes 0 minutes when state is Complete or Power off. Active states continue to use the wire value unchanged.

## Smart-course behavior and restarts

A zero smart-course ID in an idle/off frame means "not currently running", not "nothing installed". The handler no longer publishes `None` or causes Home Assistant to show Unknown from that frame. While a smart course runs, `course` reports the `Downloaded Course` placeholder like the washer and dryer, with the real name on `smart_course`.

The installed/selected smart course is remembered in process and in `/app/data/rethink-course-memory.json`. A real rethink restart restores and republishes it. This matches the F24VDD washer convention: the installed download stays visible whether it is running or not.

The bridge also observes matching app-issued download frames, so choosing a downloadable course in LG's app updates Home Assistant in the same way as choosing it through HA.

## Safety boundaries

- No remote power-on control because LG's app does not provide one and no frame was captured.
- Remote start and child lock are read-only because the appliance exposes state but no matching remote toggle command.
- Unknown non-zero error codes remain visible as `Code N` rather than borrowing labels from another model.
- S5BBP-only options are not exposed.

## Recent updates

- Energy sensor renamed from Power to Energy (Wh, `total_increasing`).
- `pause_course`/`resume_course` renamed to `pause`/`resume`.
- SmartCourse-only base course IDs labelled in COURSE so an active smart run always resolves a name.
- Memory file is no longer rewritten when the entry is unchanged.
- `reserve_hours` 1-2h rejected, matching LG's declared 0-or-3..19 range.
- Course memory is saved atomically (temp file, fsync, rename) so a crash mid-write can no longer leave a torn file; torn files from older versions still boot with defaults.
- Energy is a plain sensor again (no `entity_category`), matching the washer and the other Wh handlers in the repo.

## Verification

- `npm run check`: pass
- S5MPC tests: **28/28 pass**
- Full branch suite: **626/626 pass**
- Prettier: pass
- Deployed locally against the physical Styler through `rethink-local:latest`

## Files

- `cloud/devices/S5MPC.ts`
- `tests/cloud/devices/S5MPC.test.ts`
- `cloud/ha_bridge.ts`
- `README.md`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ node_modules
 dist
 oauth.json
 state
+captures/
--- a/README.md
+++ b/README.md
@@ -46,6 +46,7 @@ The following appliances are currently supported in rethink:
     - 👍 HCED3015D (STUDIO_HOOD), Generic identifier and probably works with multiple models. Working.
 - Stylers:
     - 👍 S5BBP (ST_B_E4H01Y_APL), Styler - mostly working
+    - 👍 S5MPC, Styler - course start / reserve / smart courses, remote-start state, error state, smart-diagnosis state, plus Power off
 
 The supported appliances can be used "out of the box" with HomeAssistant or another compatible MQTT consumer.  
 Appliances not listed above can still be used with the bridge mode, but they will not be translated to MQTT. Contributions are welcome!
--- a/cloud/devices/S5MPC.ts
+++ b/cloud/devices/S5MPC.ts
@@ -2,7 +2,7 @@ import { Device as Thinq2Device } from '../thinq2/device'
 import log from '@/util/logging'
 import { type Connection } from '../homeassistant'
 import { type Metadata } from '../thinq'
-import { readFileSync, writeFileSync } from 'node:fs'
+import { readFileSync, openSync, writeSync, fsyncSync, closeSync, renameSync, unlinkSync } from 'node:fs'
 import { allowExtendedType } from '@/util/casting'
 import HADevice from './base'
 import AABBDevice from './aabb_device'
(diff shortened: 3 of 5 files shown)


dohun0310:
Added `child_lock` (binary_sensor, read-only) in dea829c.

**Evidence**: LG's own app has no control for child lock on this appliance — it's a panel-only button — so there is no cloud snapshot to cross-check against, unlike every other field in this driver. Verified instead against the owner physically toggling the panel button and confirming the lock icon state, captured live 2026-09-08:

```
ON:  ...20 00 00 00 00 00 00 00 00 01 63 42 00 -> 00 1b 01 00 01 00 01 00 00 00 00 00 00 00 00 00 21 00 00 00 00 00 00 00 00 01 63 42 00
OFF: ...21 ... -> ... 20 ...
```

Only `flagsA` (record byte offset 14) bit `0x01` moved between the two frames; every other byte, including the existing `0x08` remote-start bit, was unchanged. Exposed read-only for the same reason `remote_start` is: the appliance takes no command for it, so a switch/lock entity would just publish a write nobody can honour.

Added the two real frames as fixtures (`CHILD_LOCK_ON`/`CHILD_LOCK_OFF`) plus an independence test against the remote-start bit. `tsc --noEmit` clean, S5MPC suite 13/13, full suite 479/479.

anszom:
You may not be aware of this, but I get an email notification each time you push to the branch. Please use a separate branch for development and push only when you have something significant.

dohun0310:
Sorry for spamming your inbox, I had no idea every push sends an email. My bad. I'll do my dev work on a separate branch from now on and only push here when there's something actually worth reviewing.

dohun0310:
I pushed one consolidated Styler update and rewrote the PR description so it matches the current branch.

The important changes are:

- `record[17..18]` is now confirmed as this-cycle energy in Wh, not instantaneous watts. The completed cycle produced `0x0251` (593), exactly matching the ThinQ app's 593Wh daily reading.
- Remaining time now reads 0 after completion/power-off instead of holding the appliance's stale 1-minute value.
- Smart course follows the washer behavior: the installed download stays visible while idle, survives a rethink restart, and app-issued downloads update HA as well.

This was kept on a separate development branch until it was ready, then merged here as a single reviewable batch. S5MPC passes 23/23 and the full branch suite passes 621/621.

dohun0310:
Pushed 8 commits since the last update:

- `2ec754d` Energy sensor renamed from Power to Energy (Wh, `total_increasing`)
- `799d89d` Completion/energy fixtures added to the intact-corpus check
- `d0be082` `pause_course`/`resume_course` renamed to `pause`/`resume`
- `fb3cd05` SmartCourse-only base course IDs labelled in COURSE
- `82fe6f8` Memory file no longer rewritten when the entry is unchanged
- `13bef63` `reserve_hours` 1-2h rejected, matching LG's declared 0-or-3..19 range
- `2c7f63b` `course` reports Downloaded Course while a smart run is active, like the washer and dryer
- `ce038cb` Incoming AABB checksum verified on receive, plus atomic course-memory saves (temp file, fsync, rename); torn files from older versions still boot with defaults

Verification: `npm run check` pass, S5MPC tests 28/28, full branch suite 626/626, Prettier pass. Deployed locally as `rethink-local:latest` and confirmed live against the physical Styler (provisioned, discovery published, real packets flowing, no errors).

dohun0310:
Pushed 1 commit since the last update:

- `6c239b8` Energy is a plain sensor again (no `entity_category`), matching the washer and every other Wh handler in the repo, so it stays on the device card instead of under Diagnostics. Pinned by a discovery assertion.

Verification: `npm run check` pass, S5MPC tests 28/28, full branch suite 626/626, Prettier pass. Deployed locally as `rethink-local:latest`, Styler online with Energy published without `entity_category`.

PR #175, opened 2026-09-07
T1789EFH_F, RV13U6AM8W_D_US_WIFI: complete both handlers

D34DC3N73R:
Rewrites the two handlers for my LG WT7300CW washer and DLG7301WE dryer. Both were listed as
preliminary; every field is now confirmed against the appliances rather than inferred.

### Decode fixes

- **`0xEC` frames were read from the wrong record.** They carry the OLD state first and the current
  one last — across consecutive captures each frame's first record is byte-identical to the previous
  frame's second — so every value was published one update late.
- **The remaining-time decode ignored the hour byte**, reporting a captured 2h06 cycle as 6 minutes.
- **The washer's status table was misaligned with the model JSON's state enum.** Index 8 is END but
  was published as "Spin", so a finished wash never reported a completed state. The `preState` byte
  walks the cycle in exactly the order the corrected table names, which is what pins the alignment.
- **Frames the exact-length checks dropped are now decoded**, including the three-record frame a
  mid-cycle power-off produces — previously the dryer stayed "Drying" until the next reconnect.

### Corrections that also affect the existing dryers

Both found by toggling the buttons on the appliance, so worth a look on RV13B6BSD/RV13B6ES too:

- **`rec[16]` bit `0x01` is Remote Start, not the drum.** It sets while the dryer sits idle in
  `Initial`, which no drum bit can do. It is *also* set through every Drying frame — not because those
  captures happened to be started remotely, but because the appliance opens a remote-control session
  for the length of any run, including one started at the panel. It tracks that session rather than
  arming: on one load an app pause held the bit and a panel pause ten minutes later cleared it, and it
  returned when the panel resumed the run. It reads set through Cooling, when a gas dryer's burner is
  off by definition, which retires the heater reading along with the drum one. RV13B6ES already had
  the identification right. The clearest frame pair is a Bedding cycle started **at the panel** and
  never armed by hand: the bit reads clear as Drying begins and set one second later, with nobody
  touching the machine. The washer does the opposite — 63 status frames of a panel-started Normal
  wash, `rec[17]` clear in every one — so this is appliance-specific behaviour, not a family
  convention.
- **`rec[15]` bit `0x40` is the drum light.** Pressing the button sets it, pressing again clears it,
  and it clears itself when the lamp times out. RV13B6BSD describes it as "present while
  Initial/Selecting/Pause, clear while actively drying" — which is what a drum light does.
- **Wrinkle Care is at `rec[15]` bit `0x10`** here, matching RV13B6ES. The two siblings disagree and
  taking the wrong one publishes it as permanently OFF.

### Added

Course tables (11 washer, 17 dryer — the dryer's includes 9 base cycles reachable only via a smart
course), 22 and 20 smart-course codes, the option bits, both washer locks, previous state, and
start/pause/resume/power-off buttons. The start commands are the first in the project; they are
byte-identical to what the LG cloud sends, captured in bridge mode, and built from the appliance's own
reported state so they run whatever is set on the machine.

The dryer's `0xF026` carries its verb as a bit inside offset 10 rather than as a trailing byte like the
washer's: every captured start reads `0x41` there and the resume the LG app sent into a live pause
reads `0x01`, so pressing Start while the record says Paused resumes at the remaining time instead of
re-running the cycle from zero. Neither Start button is gated on Remote Start being armed — both
appliances refuse an unarmed start themselves, and on the dryer the bit is clear for a cycle paused at
the panel, which is exactly when someone wants to resume.

`RV13U6AM8W_D_US_WIFI` covers both the DLE7300WE and the gas DLG7301WE, whose frames decode
identically — nothing in the handler is fuel-specific. The JSON's `EnergyMonitoring` powertable models
a resistive element and is meaningless on the gas variant, so it is unused.

### Method and testing

Every course code, smart-course code and option bit was isolated one at a time on the appliance with
the control panel or the LG app naming it, and cross-checked against the per-course defaults the model
JSON declares.

Some fields are now understood but still deliberately unpublished. The washer's `rec[11]` is a rinse
**setting**, not a live count: pressing Extra Rinse moved it 0 → 2 with the estimate 60 → 84 minutes,
starting the cycle took it to 4, and it then held at 4 through all four rinses of a Bedding wash without
decrementing — but its unit is only meaningful combined with the course's own default, which is not
separable here. The dryer's elapsed run time is a 16-bit value at `rec[18:20]`, high byte first (a
79-minute load carried `rec[19]` 212 → 21 while `rec[18]` stepped 0 → 1, so it does not wrap at 63.75
minutes as a single byte would) — not published because `initial_time − remaining_time` is the same
number from two entities that already exist.

Others are not published because they turned out not to be what they looked like: `rec[22]` never moved
through two complete washes, so it is not the tub-clean counter. Load item and three `rec[16]` bits that
are set in every frame ever captured remain unidentified.

All fixtures are captured frames, not synthesised buffers. Command tests assert byte equality with the
frames the LG cloud actually sent. 520 tests pass; each commit builds and tests standing alone.

Written with an LLM agent driving the analysis, with me operating the appliances and confirming each
mapping against the panel. Happy to adjust anything that does not fit the project's conventions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ The following appliances are currently supported in rethink:
     - 🫤 F4WV709P1E, Front-Loading Washing Machine - preliminary support
     - 🫤 TW4V9RW9W - preliminary support
     - 👍 F4X7511TWS (VCDWL2QEUK), Front-Load Washing Machine - mostly working
-    - 🫤 WT7300CW - preliminary support
+    - 👍 WT7300CW (T1789EFH_F), Top-Load Washing Machine - mostly working
     - 👍 WM3900HBA (F3L2CYU\_\_), Front-Load Washing Machine - mostly working
     - 👍 FV1413H2B, Washing Machine - mostly working,
     - 👍 F3L7CYK5W_US_WIFI, Front-Load Washing Machine - mostly working
--- a/cloud/devices/T1789EFH_F.ts
+++ b/cloud/devices/T1789EFH_F.ts
@@ -6,14 +6,271 @@ import { allowExtendedType } from '@/util/casting'
 import AABBDevice from './aabb_device'
 import { Enum } from '@/util/enum'
 
+// LG WT7300CW top-load washer — matched on modelId "T1789EFH_F". AABB frames (buf = the AABB body,
+// AA+len and checksum+BB already stripped) start with 0x20 and are discriminated by buf[1]:
+//   0xEC  status frame — stacked records, each led by a 0x19 marker: a 27-byte OLD state followed by
+//         the 26-byte CURRENT state. The current record is always the LAST one in the frame.
+//   0xEB  single-record status frame (26-byte record at buf[3:]), sent in reply to a status query and
+//         right after the appliance reconnects.
+//   0xE2  settings echo — NOT decoded. It has the same shape as 0xEB with a valid 0x19 record, but the
+//         sibling F3L7CYK5W_US_WIFI found its equivalent replays a stale snapshot of the cycle's start;
+//         decoding it would knock the machine back a phase. The frame-type gate below is what keeps it
+//         out, so it must stay even though the record itself would parse.
+//   0x31  one-time device-ID/serial frame at connect, 0xD8 short heartbeat — not decoded.
+//
+// Field offsets follow the layout shared by the other US washers and dryers in this directory (see
+// F3L7CYK5W_US_WIFI.ts and RV13B6ES_D_US_WIFI.ts), and each one below is confirmed against captured
+// frames from this appliance: the value read at the offset matches what the model JSON declares as the
+// default for the course that was running in that capture, and the phase/preState pair walks the cycle
+// in the right order across the pause/resume/spin/end captures.
+
+const RECORD_MARKER = 0x19
+const RECORD_LEN = 26
+const HEADER_LEN = 3
+const STATUS_FRAME_TYPES = [0xeb, 0xec]
+
+// Status query, sent on every connect. 0xF0ED is the family-wide "report your state" request — the
+// fridges, the EU washers, the US dryers and the WashTower all use it, and actuating commands are
+// 0xF0E5, so this only ever reads. The real LG cloud sends this same request to this washer in bridge
+// mode and it answers with a 0xEB snapshot; asking ourselves is what makes the entities survive a
+// rethink restart instead of staying blank until the next physical interaction.
+const STATUS_REQUEST = 'F0ED1121010000001800'
+
+// Offsets below are relative to the record's own 0x19 marker (rec[0]).
+const PHASE_OFFSET = 1
+// rec[2:4] = [hour][minute], the live countdown. The hour byte is not decoration: a captured 0xE2
+// settings frame carries 01 02 — 62 minutes, which reading the minute byte alone reports as 2.
+const TIME_HOUR_OFFSET = 2
+const TIME_MIN_OFFSET = 3
+// rec[4:6] = [hour][minute], the cycle's total estimate. It pins once a cycle starts while rec[2:4]
+// counts down (43-of-52 minutes remaining mid-Bedding, 24-of-86 mid-Heavy Duty).
+const INITIAL_TIME_HOUR_OFFSET = 4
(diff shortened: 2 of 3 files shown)


D34DC3N73R:
Pushed an update after another capture session on the dryer, plus a small follow-up on the washer.

**Resume.** `0xF026` does carry a verb after all — it is a bit inside offset 10, not a trailing byte. Every captured start reads `0x41` there; the resume the LG app sent into a live pause reads `0x01`, and the cycle picked up at its remaining 29 minutes rather than restarting. Pressing Start while the record says Paused now sends that, asserted byte-identical against the app's own frame.

**Dropped the `availability` gate on both Start buttons** (separate commit for the washer). It is a second interlock on top of the one the appliances already enforce, which CONTRIBUTING rules out, and on the dryer it misfired: a cycle paused *at the panel* reports Remote Start OFF, so the button was hidden precisely when someone standing at the machine wanted to resume. `remote_start` stays published as its own entity for anyone who wants to condition on it.

**Corrected the Remote Start reasoning in the description above.** The bit is a remote-control session, not arming: the dryer opens one for any run, an app pause holds it, a panel pause clears it, and it returns when the panel resumes. It also reads set through Cooling — the burner is off there on a gas dryer, so that retires the old heater theory as well as the drum one.

**New fixtures, all from one load recorded end to end:** app pause, panel pause, Cooling, End, and Wrinkle Care `0x38` — the first capture of that phase from this appliance, which had been mapped from the sibling handlers alone. Opening the door during the tumble ends it straight to Off with `pre_state` `0x38`, and nothing in the record marks the door, matching the model JSON declaring no `doorClose`.

Also added a real `0xE2` idle snapshot to the ignore test. It is `0x1b`-marked and exactly record-length, so it passes both the marker and length checks and only the frame-type filter stops it — decoded, its record would have published a fresh 60-minute Drying cycle on a dryer standing open with a finished load in it.

518 tests pass; each commit still builds and tests standing alone.

D34DC3N73R:
Pushed again after two more live loads — a Bedding wash and two dryer cycles, start to finish. No new entities; this is the last of the unknowns being closed or deliberately left closed. The two commits are amended rather than stacked, so the PR is still three commits that each build and test alone.

**Elapsed run time carries, it does not wrap.** This was the one thing blocking a decision. A 79-minute load crossed the boundary: `rec[19]` went 212 → 21 while `rec[18]` stepped 0 → 1, so `rec[18:20]` is a 16-bit counter in 15-second units, high byte first — the same most-significant-first convention as the `[hour][minute]` fields. It ended at 311 = 77.75 min. Still unpublished, but now by choice rather than doubt.

**`initial_time` is not a fixed estimate.** It tracks `remaining + elapsed`, recomputed live — Sensor Dry revised a load upward mid-run and both the countdown and the total jumped by the same 19 minutes in one frame. The two halves round independently, so consecutive frames flutter by a minute. That relationship is also what cross-checks the elapsed decode at every frame of both runs, independently of counting them.

**The remote-start bit, on the cleanest evidence yet.** A Bedding cycle started at the panel and never armed by hand reads `rec[16] & 0x01` clear as Drying begins and set one second later, with nobody touching the machine — the appliance opening a remote-control session for itself. The washer is the contrast that makes it appliance-specific rather than a family convention: 63 status frames of a panel-started Normal wash, `rec[17]` clear in every one.

**Washer, from two complete cycles watched end to end.** `rec[11]` is a rinse setting rather than a live count (held at 4 through all four rinses). A cycle reaching Complete disarms Remote Start itself and unlocks the lid a second behind it, and Complete lasts only ~30 seconds before the washer drops to Off. `rec[22]` never moved across both washes, so it is not the tub-clean counter — a negative result, but it is the difference between an unpublished field and a sensor pinned at zero.

**The `0xE2` trap is now first-hand on both appliances** rather than inherited from `F3L7CYK5W_US_WIFI`. Four arrived three seconds after a wash finished, carrying the state from ninety minutes earlier (Washing, 1h42 remaining); the dryer's equivalent replays its cycle's start state after the load is done. They are `0x19`/`0x1b`-marked and exactly record-length, so the frame-type gate is the only thing stopping them — one of each is now a fixture.

520 tests pass.

PR #179, opened 2026-09-09
Fix WIN_056905_WW.ts protocol mapping and add filter/timer/power support

BatemanJo9:
Confirmed against a real LW1822IVSM:
- dry mode's raw value was wrong (was 8, actually Energy Saver; corrected to 1)
- medium fan speed was missing from the fan table
- Energy Saver now exposed as a preset_mode
- filter usage tracking added (used/life/last-reset + reset button), ported from RAC_056905_WW.ts's private-command protocol
- live power draw sensor added (same tag/formula as RAC's 'energy_current')
- off timer exposed as a number entity, write path confirmed live
- fixed a setProperty('power', ...) typo that silently no-op'd
- fixed the same 0x87-only framing gap in util/packet-codec.ts that tlv_device.ts already had a fix for
- README/docstring updated to list LW1822IVSM as its own supported model

### DIFF
--- a/cloud/devices/WIN_056905_WW.ts
+++ b/cloud/devices/WIN_056905_WW.ts
@@ -52,9 +52,25 @@ export default class Device extends TLVDevice {
                     platform: 'climate',
                     unique_id: '$deviceid-climate',
                     name: null,
-                    temperature_unit: 'C',
-                    temp_step: 0.5,
-                    precision: 0.5,
+                    // Native unit is deliberately Fahrenheit, not the wire protocol's Celsius. This
+                    // unit only ever displays/accepts whole-degree Fahrenheit (61-86F, see the
+                    // temperature field's clamp below); if the entity's native unit were 'C' with a
+                    // 0.5 step, an HA instance running an imperial unit system would round-trip every
+                    // setpoint through an unaligned Celsius grid (e.g. typing 74F snaps to 23.5C,
+                    // which displays back as 74.5F, while the AC's own firmware converts that same
+                    // 23.5C to 75F on its panel) — confirmed live (2026-09-09). Declaring 'F' here
+                    // means HA never does that conversion; write_xform/read_xform below do the
+                    // C<->F math once, explicitly, against the wire's Celsius-based raw value.
+                    temperature_unit: 'F',
+                    temp_step: 1,
+                    precision: 1,
+                    // HA's MQTT climate min_temp/max_temp default to 7/35 -- correct for the old
+                    // Celsius-native config (bracketing 16-30C) but silently wrong now that the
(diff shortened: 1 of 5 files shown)


cloud/devices/WIN_056905_WW.ts
```diff
@@ -18,25 +19,86 @@ const SWING_MODES = Enum.of({
...
+                    // Native unit is deliberately Fahrenheit, not the wire protocol's Celsius. This
+                    // unit only ever displays/accepts whole-degree Fahrenheit (61-86F, see the
+                    // temperature field's clamp below); if the entity's native unit were 'C' with a
+                    // 0.5 step, an HA instance running an imperial unit system would round-trip every
+                    // setpoint through an unaligned Celsius grid (e.g. typing 74F snaps to 23.5C,
+                    // which displays back as 74.5F, while the AC's own firmware converts that same
+                    // 23.5C to 75F on its panel) — confirmed live (2026-09-09). Declaring 'F' here
+                    // means HA never does that conversion; write_xform/read_xform below do the
+                    // C<->F math once, explicitly, against the wire's Celsius-based raw value.
+                    temperature_unit: 'F',
```
maciejsszmigiero, line 64: Are you sure that this ThinQ model name (`WIN_056905_WW`) is only used and will only be used for AC units sold in countries using the Fahrenheit scale?

The `_WW` suffix would suggest a World Wide device.

cloud/devices/WIN_056905_WW.ts
```diff
@@ -47,22 +109,23 @@ export default class Device extends TLVDevice {
             comp: 'climate',
             state_topic: 'topic',
             writable: false,
-            read_xform: (raw) => raw / 2,
+            // raw is Celsius*2 on the wire; convert to whole Fahrenheit to match the entity's
+            // native unit (see the climate component's temperature_unit comment above).
+            read_xform: (raw) => Math.round((raw / 2) * (9 / 5) + 32),
```
maciejsszmigiero, line 114: Was this mapping checked across the whole temperature range?

AFAIK LG devices operate in Celsius internally but when set to the Fahrenheit mode convert C <-> F in some non-standard way, see [here](https://github.com/JanM321/esphome-lg-controller/issues/27).

cloud/devices/WIN_056905_WW.ts
BatemanJo9: Sorry, I pushed another fix. I assumed this unit was Fahrenheit since the font panel has an embedded F next to the display and the manual does not show you can change the display to Celsius. I found you can change it to C. Also my issue was deep inside HA doing the wrong conversion.

cloud/devices/WIN_056905_WW.ts
BatemanJo9: I fixed the mapping in Home Assistant instead

PR #180, opened 2026-09-10
Add F24VDD washer support

dohun0310:
## Summary

Adds first-class support for the **LG F24VDD front-load washer** (`deviceType: 201`). This is a dedicated AA/BB handler built from captures from the physical appliance rather than an alias of another washer.

It supports the washer's native and downloadable courses, course-specific options, remote controls that were actually captured, Home Assistant state reporting, and this-cycle energy.

## What is supported

- 14 captured native course start templates with per-course rinse, spin, and temperature restrictions
- All 14 downloadable courses, including install, start, reservation, and resume
- `course_select` and `smart_course_select`, with `Downloaded Course` routing
- Start, pause, resume, remote power off, and completion reservation (0 or 3 to 19 hours; 1-2h rejected per LG's declared range)
- Power, status, course, smart course, temperature, spin, rinse, TurboShot, steam, remote start, child lock, errors, diagnosis, and duration sensors
- This-cycle energy in Wh

The downloadable-course IDs were captured individually. They are not inferred from list position. `Spin Only`, for example, reports ID 63 rather than the value an index-based mapping would predict.

## Energy

The washer's normal 36-byte status record does not contain the energy value. It arrives in the larger detailed telemetry frames:

- `0xcd`, 405-byte body: `buf[36]`
- `0xbd`, 406 or 476-byte body: `buf[37]`

The field is cumulative energy for the current cycle in Wh, with a multiplier of 1. During a complete wash it rose from 5 to 221, and the ThinQ app showed exactly 221Wh for that day. The old placeholder zero has been removed.

The appliance only sends these detailed frames during a cycle, so after a restart the sensor may retain its previous value until the next detailed frame arrives.

## Smart-course behavior

The appliance reports its installed downloadable-course ID in every normal status frame. `smart_course` and both selects therefore recover directly from the appliance after reconnecting and stay synchronized whether the downloadable course is currently running or not.

## Safety boundaries

- No remote power-on command is exposed because it was not captured.
- Child lock remains read-only because it is a panel-only operation.
- TurboShot and steam are read-only too: the options are set on the panel or in the app, the wire only reports them.
- Writes outside a selected course's captured option range are ignored rather than clamped or guessed.
- Downloadable resume uses the same start-flag transition captured on both a native course and Cold Wash. This shared mechanism is applied to the remaining downloadable templates.

## Recent updates

- Out-of-range spin codes report as `Code N` instead of folding into Off (spin 0 stays Off; it is a real setting).
- `reserve_hours` 1-2h is rejected, matching LG's declared 0-or-3..19 range.
- TurboShot reported as a read-only sensor (bit `0x80` of record byte 16), confirmed from an owner-labelled ON/OFF pair on the same 16-hour reservation.
- Steam reported as a read-only sensor (bit `0x10` of the flags byte), confirmed from an owner-labelled steam-ON run against the same-settings steam-off frame.

## Evidence

The test fixtures include real single and double status records, all native and downloadable command frames, child-lock transitions, TurboShot and steam transitions, reserved starts, pause/resume, and detailed energy telemetry. AA/BB lengths and checksums are validated in the tests.

## Verification

- `npm run check`: pass
- F24VDD tests: **95/95 pass**
- Full branch suite: **693/693 pass**
- Prettier: pass
- Deployed locally against the physical washer through `rethink-local:latest`

## Files

- `cloud/devices/F24VDD.ts`
- `tests/cloud/devices/F24VDD.test.ts`
- `cloud/ha_bridge.ts`
- `README.md`


dohun0310:
I checked the latest washer changes against the PR branch and updated the description to match what is actually implemented now.

The main change since the original write-up is energy monitoring. The detailed telemetry frames carry a per-cycle Wh counter (`0xcd` at `buf[36]`, `0xbd` at `buf[37]`). A full cycle ended at 221 and the ThinQ app showed 221Wh for the same day, so the placeholder zero is gone and the sensor now publishes the captured value directly.

I also reran the branch after the documentation update: the F24VDD tests pass 90/90 and the full suite passes 688/688.

dohun0310:
Pushed 3 commits since the last update:

- `0b1e488` Out-of-range spin codes report as `Code N`, not Off (spin 0 is a real setting and still reads Off)
- `91ebfc6` `reserve_hours` 1-2h rejected, matching LG's declared 0-or-3..19 range
- `ea5e7ce` Incoming AABB checksum verified on receive; corrupted frames are dropped instead of decoded (scoped to this device, with a test that flips one byte of CURRENT)

Verification: `npm run check` pass, F24VDD tests 93/93, full branch suite 691/691, Prettier pass. Deployed locally as `rethink-local:latest` against the physical washer.

dohun0310:
Pushed 2 commits since the last update, both from live owner-labelled captures tonight:

- `d811ed7` TurboShot as a read-only sensor (bit `0x80` of record byte 16). Same 16-hour Standard/60C/rinse-1/Extra-low reservation with TurboShot ON then OFF; the only non-clock difference in all 36 bytes was that one bit.
- `5c0acfd` Steam as a read-only sensor (bit `0x10` of the flags byte). Same-settings steam-ON run vs steam-off; temp-Off alone does not set the bit. Also explains the earlier `0x30` flags on the steam courses.

Both follow the dryer steam convention (plain read-only `binary_sensor`, no command, no `entity_category`). Verification: `npm run check` pass, F24VDD tests 95/95, full branch suite 693/693, Prettier pass. Deployed locally as `rethink-local:latest`; TurboShot and Steam discovery verified live on the physical washer.

Issue #70, opened 2026-06-09
RAC_056905_WW with RTOS 2.6.7 firmware: setup script unsupported, runtime falls into "unknown" branch


erhankeseli:
Device

  - LG split AC, model PC12SQ NSJ (S3NM12JA2FA), DualCool Standard Plus, made in Turkey 2024
  - WiFi module hostname (DHCP): QCA4002 — Qualcomm Atheros chip
  - MAC OUI 2c:2b:f9 (LG Innotek)
  - modelId reported by device: RAC_056905_WW
  - DevInfo payload (decoded): TimeZone=+0100,GroupType=1,FwVer=2.6.7_RTOS_3K,regFail=N

  So the modelId matches the existing class but the firmware string 2.6.7_RTOS_3K looks like a different platform than what cloud/devices/RAC_056905_WW.ts was written for.

  Problem 1 — rethink-setup.ts does not work against this device

  After putting the unit into AP mode (remote: SWING/SET UP for 3-6s, AP indicator on display, SoftAP LGE_AC appears, DHCP hands out 192.168.1.2, gateway 192.168.1.1), running the setup script fails.

  What the script tries:
  1. Connects to port 5500 — closed on this device. Only port 9000 is open.
  2. After patching the script to use 9000, TLS handshake fails with unsafe legacy renegotiation disabled until SSL_OP_LEGACY_SERVER_CONNECT is set.
  3. After patching that, server cert is 1024-bit RSA, requires @SECLEVEL=0 cipher list.
  4. After all that, TLS finishes, the mtosp/XML framed deviceinfo packet is sent (116 bytes, magic aa006e…), and the device immediately FINs the socket without sending a single byte (hadErr=false, peer closed write).
  Same behaviour with the ThinQ2 JSON path.

  The actual setup protocol on this device is plain HTTP on port 9000, HTTP/0.9 responses, with x-lgedm-password / x-lgedm-userid headers (matches the pentestpartners 2016 write-up:
  https://www.pentestpartners.com/security-blog/the-wi-fi-enabled-lg-air-conditioner/).

  Verified manually — GET /GetApList with the headers from bridge/thinq1connection.ts triggers the unit to beep and return a binary payload:

  HTTP/1.1 (the device speaks HTTP/0.9 framing — curl needs --http0.9)
  Body (first 7 bytes captured before timeout):
  00000000: 15 00 00 00 02 02 0a

[...]

PR #185, opened 2026-09-12
Add ThinQ1 support for RAC_056905_WW air conditioners

nicky1983:
Adds ThinQ1 support for RAC_056905_WW air conditioners using the
legacy JSON-based protocol.

This model already exists in rethink as a ThinQ2/TLV device, but some
older RAC_056905_WW units use the ThinQ1 protocol instead.

Implemented and tested with two physical RAC_056905_WW units running
firmware 2.6.7_RTOS_3K.

Supported features:
- Power on/off
- HVAC mode
- Current temperature
- Target temperature
- Fan speed
- Home Assistant MQTT discovery
- Separate power switch

The ThinQ1 implementation is registered separately from the existing
ThinQ2 RAC_056905_WW implementation, so existing ThinQ2 support is left
unchanged.

Validation:
- npm run check passes
- npm test: 605 passed, 0 failed

This addresses the runtime/device-handler side of #70.
Provisioning/setup is not addressed by this PR.

### DIFF
--- a/cloud/devices/RAC_056905_WW_thinq1.ts
+++ b/cloud/devices/RAC_056905_WW_thinq1.ts
@@ -1,0 +1,240 @@
+import HADevice from './base'
+import { Device as Thinq1Device } from '../thinq1/device'
+import { type Connection } from '../homeassistant'
+import { allowExtendedType } from '@/util/casting'
+import { type Metadata } from '../thinq'
+
+const MODE_TO_CODE: Record<string, string> = {
+    cool: '0',
+    dry: '1',
+    fan_only: '2',
+    heat: '4',
+    auto: '6',
+}
+
+const CODE_TO_MODE: Record<string, string> = {
+    '0': 'cool',
+    '1': 'dry',
+    '2': 'fan_only',
+    '4': 'heat',
+    '6': 'auto',
+}
+
+const FAN_TO_CODE: Record<string, string> = {
+    'very low': '2',
+    low: '3',
+    medium: '4',
+    high: '5',
+    'very high': '6',
+    auto: '8',
+}
+
+const CODE_TO_FAN: Record<string, string> = {
+    '2': 'very low',
+    '3': 'low',
+    '4': 'medium',
+    '5': 'high',
+    '6': 'very high',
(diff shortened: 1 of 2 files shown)


BenJamesAndo:
Hi @nicky1983 , thanks for this. I have three WH09SKN-18 units (RAC_056905_WW, also 2.6.7_RTOS_3K) and reached the same conclusion: the status payload is JSON and this variant needs its own ThinQ1 handler alongside the TLV one. A few things I found while getting them fully local that may save you some time.

Control envelope. {Cmd:"Control", CmdOpt:"Set", Value:{...}} on its own is accepted by the appliance but silently ignored: no ack, no NAK, no state change. It needs Format:"B64" and Data:"" present as well, so {Cmd:"Control", CmdOpt:"Set", Value:{"TempCfg":"23"}, Format:"B64", Data:""}. I confirmed this against readback for Operation, OpMode (0/1/2/4/6), WindStrength (2 through 6 and 8), TempCfg, both swing fields, Jet and AirClean. Worth checking whether your writes are actually landing, or whether the test units behave differently.

Placeholder status. For roughly the first 18 s after Mon Start, a powered-off unit pushes a status with 78 of 81 fields set to "0" (TempCfg included) before it has read real state. Dropping packets with TempCfg == "0" avoids publishing a 0 degree setpoint to HA.

Mode-dependent rejects. TempCfg is ignored in fan_only and dry, and PowerSave and DisplayControl rejected every write I tried, so I left those two out. Jet and AirClean work (Jet pins the setpoint to 18 and fan to max, which is the appliance's own behaviour).

I have a version with a test file built from captured packets, swing on both axes, Jet and Air purify, and labels matching the TLV handler's, plus a separate change adding this variant's SoftAP provisioning (it is HTTP over TLS on port 9000, not 5500). Happy to either send the test fixtures and the envelope fix your way for this PR, or open mine as an alternative if @anszom would rather review one complete PR. Whichever is best.
I've drafted a PR as an example at [https://github.com/BenJamesAndo/rethink/pull/1](https://github.com/BenJamesAndo/rethink/pull/1)

PR #187, opened 2026-09-17
Add support for WM3600HWA (F3P2CYVAT__) front-load washer

### LINKED ISSUE #186: Add washer: WM3600HWA (F3P2CYVAT__)
Working to add WM3600HWA (F3P2CYVAT__) as a new device. 
Product code: F3P2CYUAW.ABWEUUS on the machine.
Using F3L2CYU__ as template. 
I will update with progress.

shreead:
Closes #186


Nuck-TH:
~~With all due respect, but i recommend to look at how other devices support is implemented here. In description and code - your approach to both is significantly different and that makes it confusing and... redundant. TLV packets for both capabilities and settings/measured values are already parsed and packed into arrays, so you don't need offsets and bit manipulation - just what specific tag types mean and what to make of their values.~~

Scratch that, i wasn't paying enough attention and it passed me that it is AABB device. Sorry.

